Issue #8, opened 2019-08-26
pytest 5.1.0 outputs incompatible junit.xml


ptillemans:
In pytest-5.1.0 bugfix https://github.com/pytest-dev/pytest/issues/5477  was included which adds a *testsuites* root element.

see also https://github.com/useblocks/sphinx-test-reports/issues/7

This breaks test reports in sphinx.

danwos:
Thanks for the bug report and especially for the pull request.
Merged it and uploaded a new version 0.2.1.

Added you also to the AUTHORS file. I hope that's okay for you.

And thanks again for your help!

PR #9, opened 2019-08-26
skip testsuites element if present

ptillemans:
simple workaround to fix the **testsuites** toplevel element which suddenly appeared in pytests-5.1.0


ptillemans:
fixes https://github.com/useblocks/sphinx-test-reports/issues/8

PR #18, opened 2021-08-01
Fix/junit parser skipped tests lookup

JamesMBarr:
1. Adds a fix for getting skipped tests for test suites. See [junit schema](https://github.com/windyroad/JUnit-Schema/blob/162a883ac631f61ca1d14a3d472e1fbb53c1c05f/JUnit.xsd#L196)
2. Adds support for pytest xfail test cases which does not have "text"


danwos:
Thanks for the PR and sorry that it took so long.

I needed so solve some conflicts and have done it with PR #24, which includes your commits.
I hope that's okay.

#24 got already merged. So I close this PR.

PR #31, opened 2022-05-19
Supports nested testsuites

mawieland:


sphinxcontrib/test_reports/directives/test_case.py
```diff
@@ -74,6 +86,15 @@ def run(self):
                 case = case_obj
                 break
 
+            elif nested and case_count > -1:
```
danwos, line 89: I like `case_count >= 0` more, as it directly shows you the first allowed value.
For me easier to read

danwos (changes requested):
Some smaller findings.
Overall really good work.

sphinxcontrib/test_reports/directives/test_case.py
mawieland: agreed

danwos (approved):

Issue #13, opened 2021-04-17
Custom template for the test report directive


jakehader:
It does not appear that adding a custom test_report template is supported in the `TestReportDirective` class:

https://github.com/useblocks/sphinx-test-reports/blob/20e1828a01b9085ec36e088d79cdd0b3c4f37efa/sphinxcontrib/test_reports/directives/test_report.py#L37-L38

It would be nice if this could be pulled out as a configurable setting.

danwos:
Agree, that would be really helpful.

I suggest a config var ``tr_report_template``, which needs a relative path from ``conf.py`` location.
Then this is used as template.

But I think some additional docs are needed to describe the usage of templates and the available vars in it.

jakehader:
Thanks for the reply! I forked the repo and was trying to play around with adding something to `conf.py` but didn't know where to start. I was trying to use the `sphinx-needs` project as a example but didn't get very far.

I would be happy to help and submit a PR if you could link some documentation on how you'd like it done. It might be faster for you to implement though given my limited knowledge of the project right now.

elupus:
Seems this is resolved?

PR #33, opened 2022-05-23
adding support for custom templates for testreports

mawieland:
implements https://github.com/useblocks/sphinx-test-reports/issues/13

- added custom template option in conf.py with checks for correct paths
- moved template visualisation to template file
- added documentaion about the new feature
- added InvalidConfigurationError


docs/configuration.rst
```diff
@@ -69,3 +69,9 @@ By default ``tr_case`` is set to::
 
 Please read :ref:`tr_file` for more details.
 
+tr_report_template
+------------------
+
+``tr_report_template`` allows to specify a custom template for testcase visualisation. Provide a filename and place the
+file in the same directory as your conf.py (or provide an absolute path). For an example with just a changed order uncomment
+``tr_report_template = "custom_test_report_template.txt"`` in conf.py
```
danwos, line 77: > For an example with just a changed order uncomment
``tr_report_template = "custom_test_report_template.txt"`` in conf.py

The `conf.py` is only available in the github repo, in which a user normal does not take a look.
Or in worst case, the user things ``tr_report_template`` is already set in his/her project conf.py.

If you want to reference/show an example, I would integrate the file as code-block here:
``.. literalinclude:: file.template``.

docs/configuration.rst
```diff
@@ -69,3 +69,9 @@ By default ``tr_case`` is set to::
 
 Please read :ref:`tr_file` for more details.
 
+tr_report_template
+------------------
+
+``tr_report_template`` allows to specify a custom template for testcase visualisation. Provide a filename and place the
+file in the same directory as your conf.py (or provide an absolute path). For an example with just a changed order uncomment
```
danwos, line 76: Isn't the path relative?
So could it also be ``my/sub/folder/template.rst`` (relative to the conf.py folder)?

sphinxcontrib/test_reports/directives/test_report.py
```diff
@@ -39,9 +40,26 @@ def run(self):
...
-            os.path.join(os.path.dirname(__file__), "test_report_template.txt")
-        ) as template_file:
+        # if user provides a custom template, use it
+        tr_template = self.app.config.tr_report_template
+
+        template_path = ""
+
+        if os.path.isabs(tr_template):
+            template_path = tr_template
+
```
danwos, line 50: remove newline around ``else`` :)

danwos (changes requested):
Good working.
Found smaller stuff in docs only.

I think we should add also a small test case.
Which checks, if the output is the correct one, if ``tr_report_template`` is set.

danwos (approved):
Looks all good, thanks for the PR :+1:

PR #65, opened 2023-05-16
Initial JSON Parser implementation

danwos:
* Uses JSON parser if file has json extension
* test and docs


iSOLveIT:
Hi @danwos,
While going through the code for the JSON parser, I came across the function below:
```python
        def parse_testcase(json_dict):
            # testcase = json_dict

            # ToDo: Replace dict-keys by values from conf.py
            tc_dict = {
                "classname": dict_get(json_dict, ["classname"], "unknown"),
                "file": dict_get(json_dict, ["test", "file"], "unknown"),
                "line": dict_get(json_dict, ["line"], -1),
                "name": dict_get(json_dict, ["name"], "unknown"),
                "time": dict_get(json_dict, ["time"], -1),
                "result": dict_get(json_dict, ["result"], "unknown"),
                "type": dict_get(json_dict, ["type"], "unknown"),
                "text": dict_get(json_dict, ["text"], "unknown"),
                "message": dict_get(json_dict, ["message"], "unknown"),
                "system-out": dict_get(json_dict, ["system-out"], "unknown"),
            }
            return tc_dict
```

In the function above, both the dict-keys and the selector list are all hard-coded in the function instead of using the values from conf.py. If it is ok for you I would like to work on that as part of this PR or let me know if it should be left as it is.
Thanks

danwos:
Just wanted to start working on it :)
But for sure you are welcome and can do it as well.
Thanks for the support here :+1:

PR #66, opened 2023-05-25
Correct needs types in documentation

elupus:
The documentation for needs type should match the default settings

### DIFF
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -20,9 +20,9 @@ Filterable data
 ``Sphinx-Test-Reports`` adds the following data to a sphinx-need-configuration:
 
 * **Types**
-   * test-file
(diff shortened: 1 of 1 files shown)


danwos (changes requested):
I think we should not change it, as for instance `test-file` is written like this in the rest of the documentation as well.

The only problem I see: `test_suite` and `test_case` should be `test-suite` and `test-case.

What do you think?

elupus:
I'm fine with that. But the docs elsewhere refer to these using those names. For example in test_suite.rst: "test-suite creates a need of type Test-Suite and adds the following options automatically:". Those should be changed to lowercase too in that case to make it consistent.

danwos:
Fixed it in PR #93

PR #68, opened 2023-10-30
Support pattern matching in tr_link

cuinixam:
In [our project](https://github.com/avengineers/SPLed/blob/0fcbf525c3c2d074d81615f838a99def7b311dff/src/light_controller/test/test_light_controller.cc#L231) we are using GTest value-parameterized tests - see [TEST_P](https://google.github.io/googletest/reference/testing.html#TEST_P).

**Problem:** for one parametrized test case, there are multiple test results. With the current implementation of `tr_link` we can not dynamically link all tests results to the test case itself.

**Solution:** Update `tr_link` to use pattern matching if `*` is contained in the `need` value. 

Bellow is an example of how we use parametrized tests with a copy of `tr_link` in the conf.py for our project:

```
// Define a struct to hold the parameters for each test case
struct TestParam {
    const char* description;
    percentage_t mainKnobValue;
    unsigned int expectedBlinkPeriod;
};

// Override the cout operator for TestParam so that it can be printed in the test output
std::ostream& operator<<(std::ostream& os, const TestParam& param)
{
    os << param.description;
    return os;
}

/*!
* @rst
*
* .. test:: BlinkPeriodTests/BlinkPeriodTest.CalculatesCorrectBlinkPeriod/*
*    :id: TS_LC-003
*    :results: [[tr_link('title', 'case')]]
*    :tests: SWDD_LC-002
*
* @endrst
*/
TEST_P(BlinkPeriodTest, CalculatesCorrectBlinkPeriod)
{
    // Get the parameters for this test case
    TestParam param = GetParam();

    unsigned int blinkPeriod = calculateBlinkPeriod(param.mainKnobValue);
    EXPECT_EQ(blinkPeriod, param.expectedBlinkPeriod) << "Test case: " << param.description;
}

// Instantiate the test suite with a set of parameters
INSTANTIATE_TEST_SUITE_P(
    BlinkPeriodTests,
    BlinkPeriodTest,
    ::testing::Values(
        TestParam{ "Slowest", 0, 100 },
        TestParam{ "Inbetween", 50, 50 },
        TestParam{ "Fastest", 100, 10 }
    )
);
```

The name of a test result in the generated JUnit report follows the following pattern `<test suite>/<test class>/<test case>/<test value>` and now one can match for `<test suite>/<test class>/<test case>/.*`

### DIFF
--- a/docs/functions.rst
+++ b/docs/functions.rst
@@ -10,7 +10,7 @@ from Sphinx-Needs documentation to know how to use them.
 
 
 tr_link
----------
+-------
 Links a need (e.g testcase) automatically to other needs, which have a specific value in a given option.
 
 **Usage**::
@@ -25,8 +25,8 @@ Links a need (e.g testcase) automatically to other needs, which have a specific 
 
 ``tr_link`` needs the following arguments:
 
-* **source_option**: Name of an option of the test-need, which is used for comparision. E.g. ``classname``.
-* **target_option**: Name of an option of all other needs, which is used for comparision. E.g. ``title``.
+* **source_option**: Name of an option of the test-need, which is used for comparison. E.g. ``classname``.
+* **target_option**: Name of an option of all other needs, which is used for comparison. E.g. ``title``.
 
 The function reads the ``target_option`` from the need, where it is used.
 Then it goes through **all** other needs and checks if the value of their ``source_option`` is equal to
@@ -34,7 +34,7 @@ the ``target_option``.
 If this is the case, their IDs get stored and finally returned.
 
 ``source_option`` can also reference an option with comma separated values.
-In this case a comparions is performed for each value, which may lead to multiple links.
+In this case a comparison is performed for each value, which may lead to multiple links.
 
 **Example**::
 
@@ -81,3 +81,25 @@ In this case a comparions is performed for each value, which may lead to multipl
 
 .. needflow::
    :tags: link_example
+
+
+tr_link_match
+-------------
+
+Similar to ``tr_link``, but allows using a regex to find needs to link to.  
+When using ``tr_link_match``, the value of the ``source_option`` is compiled as a regular expression, and all other needs
+whose ``target_option`` values match this pattern are linked.
+
+**Usage**::
+
+   .. test-case:: Tests/TestSuite.TestCase/*
+      :id: TESTLINK_1
+      :results: "[[tr_link_match('title', 'case')]]",
+
+The example above will link all test results which have a classname starting with ``Tests/TestSuite.TestCase`` 
+(given that the ``results`` are defined as ``needs_extra_links`` in the ``conf.py``).
+
+``tr_link_match`` needs the following arguments:
+
+* **source_option**: Name of an option of the test-need (regex is taken from its value).
+* **target_option**: Name of an option in other needs, which must match the regex in ``source_option``.
--- a/sphinxcontrib/test_reports/functions/__init__.py
+++ b/sphinxcontrib/test_reports/functions/__init__.py
@@ -1,3 +1,6 @@
+import re
+
+
 def tr_link(app, need, needs, test_option, target_option, *args, **kwargs):
     if test_option not in need:
         return ""
@@ -17,5 +20,33 @@ def tr_link(app, need, needs, test_option, target_option, *args, **kwargs):
                 and len(test_opt) > 0  # fmt: skip
             ):
                 links.append(need_target["id"])
+
+    return links
+
+
+def tr_link_match(app, need, needs, test_option, target_option, *args, **kwargs):
+    if test_option not in need:
+        return ""
+    test_option_value = need[test_option]
+    if test_option_value is None or len(test_option_value) <= 0:
+        return []
+
+    links = []
+    test_pattern = re.compile(test_option_value)
+    for need_target in needs.values():
+        # Skip linking to itself
+        if need_target["id"] == need["id"]:
+            continue
+
+        if target_option not in need_target:
+            continue
+
+        target_option_value = need_target[target_option]
+        if (
+            target_option_value is not None
+            and len(target_option_value) > 0
+            and test_pattern.match(target_option_value)
+        ):
+            links.append(need_target["id"])
 
     return links
--- a/tests/test_tr_link.py
+++ b/tests/test_tr_link.py
@@ -1,0 +1,65 @@
+from sphinxcontrib.test_reports.functions import tr_link, tr_link_match
+
+
+def test_tr_link_option_not_in_need():
+    """
+    Return an empty string when the specified test option is missing from the need.
+    """
+    assert tr_link(app=None, need={}, needs={}, test_option="a", target_option="b") == ""
+
+def test_tr_link_no_target_option_in_needs():
+    """
+    Return an empty list when the target option is missing in all items of needs.
+    """
+    assert tr_link(app=None, need={"a": "1"}, needs={"x": {"id": "123"}}, test_option="a", target_option="b") == []
+
+def test_tr_link_no_match():
+    """
+    Returns an empty list when no matching value for the test option is found in any of the target options within needs.
+    """
+    assert tr_link(app=None, need={"a": "1"}, needs={"x": {"b": "2", "id": "123"}}, test_option="a", target_option="b") == []
+
+def test_tr_link_match():
+    """
+    Returns a list of ids when there is a matching value in both need and needs.
+    """
+    assert tr_link(app=None, need={"a": "1"}, needs={"x": {"b": "1", "id": "123"}}, test_option="a", target_option="b") == ["123"]
+
+def test_tr_link_regex_match():
+    """
+    Returns a list of ids when the test option value containing an asterisk (*)
+    correctly matches target options using regular expression patterns.
+    """
+    needs = {
+        "x": {"b": "abc123", "id": "111"},
+        "q": {"b": "abc/123", "id": "112"},
+        "y": {"b": "def456", "id": "222"},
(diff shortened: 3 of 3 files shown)


tests/test_tr_link.py
```diff
@@ -0,0 +1,139 @@
...
+    spec = importlib.util.spec_from_file_location(name, path)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+# (!) We need to import the 'tr_link' method from the local module,
+# otherwise the method from the installed 'sphinxcontrib' package will be used.
+local_functions_module = import_from_path(
```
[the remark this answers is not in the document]
danwos: On my machine a `from sphinxcontrib.test_reports.functions import tr_link` was enough, after I did:

```
pip install .
pip install -r test-requirements.txt
pip install sphinxcontrib-plantuml  # a missing dep in the sphinx-needs config
```

So I guess we do not need this kind of import here.

danwos (changes requested):
Great work and thanks for the tests. :+1:
Maybe we should also add a small hint in the docs as well, that `*` is supported.

I only have some concerns with the `tr_link` import in the tests, but see my comments.

And for sure feel free to add yourself to the AUTHORS file, if you like.

tests/test_tr_link.py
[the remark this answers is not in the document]
cuinixam: I am using a virtual environment.
I did following:
* create `.venv` directory in the repository
* run `pipenv install -r test-requirements.txt`

Executing the tests with `pytest` fail because `sphinx` is not installed. So the dependencies in the `test-requirements.txt` are not enough to execute the tests.

I run `pipenv install -r doc-requirements.txt`.
All tests now pass.

I removed the ugly `import_from_path ` and just added `from sphinxcontrib.test_reports.functions import tr_link`. 
Now the `test_tr_link.py` tests are failing because the `tr_link` from the `sphinxcontrib` is used. I raised an exception in the local `tr_link` to make sure this is actually not used. The exception was not raised and the tests are failing.

So my problem is caused by installing the doc dependencies as these contain `sphinx-test-reports>=0.3.3`.

I am not familiar with `nox` so I did check why this is not a problem on CI.

There are two things to clarify:

* which steps a developer shall do locally to have a running test environment
* how to make sure that the correct `tr_link` method is used also when `sphinxcontrib.test_reports` is available in the Python environment.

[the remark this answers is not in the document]
danwos: Thanks for the insights into your setup, I guess `pip install .` or similar is missing to install the locally checked-out Sphinx-Test-Report package. Then the correct `tr_link` should be taken for tests.

cuinixam:
I have removed the ugly local import hack. Installing the local package with `pip install .` does the job. I trust that the CI setup will also use the locally checked-out Sphinx-Test-Report package for testing😉 
I also briefly mentioned this feature in the docs.

sphinxcontrib/test_reports/functions/__init__.py
```diff
@@ -1,16 +1,28 @@
...
-        ):
-            links.append(need_target["id"])
+        if test_option_value is not None and len(test_option_value) > 0:
+            target_option_value = need_target[target_option]
+            if target_option_value is not None and len(target_option_value) > 0:
+                if test_option_value == target_option_value:
+                    links.append(need_target["id"])
+                # if the test option value has a *, use regex matching
+                elif "*" in test_option_value:
+                    if re.match(test_option_value, target_option_value):
```
danwos, line 25: AFAIK `re.match` compiles the pattern with each call (pattern here is `test_option_value`)
As `test_option_value` is not changed during the for loop, we should maybe compile it in advance, near to the line `test_option_value = need[test_option]`:
`test_pattern = re.compile(test_option_value)`

Then we also could call `test_pattern.match(target_option_value)` always, not only if a `*` is in it.
Will save some performance :)

danwos (changes requested):

sphinxcontrib/test_reports/functions/__init__.py
cuinixam: Nice finding 😀

cuinixam:
I have an issue with getting the pre-commit check to pass. Flake8 complains about the line breaks before binary operators - see [W503](https://www.flake8rules.com/rules/W503.html). In the same time, black formats the code to have the line breaks before binary operators, which I actually prefer. Should I add `W503` to the list of ignored warnings in the `.flake8` file?

cuinixam:
The pre-commit check passes now on my machine 😀

cuinixam:
Any chance to review pr?

docs/functions.rst
```diff
@@ -33,6 +33,10 @@ Then it goes through **all** other needs and checks if the value of their ``sour
 the ``target_option``.
 If this is the case, their IDs get stored and finally returned.
 
+In case the ``source_option`` contains a ``*`` (e.g. ``sphinxcontrib.test_reports.*``), the function will check if the
+``target_option`` matches the ``source_option``. This is useful if one wants to link multiple test results to one test case.
+
```
danwos, line 38: I'm not sure, if this documentation is correct.
`target_option` is now used as regex-pattern in all cases and `sphinxcontrib.test_reports.*` works, because of `.*`, which allows everyhing else after the given string, so also ´sphinxcontrib.test_reportsTHIS_IS_NEW`.

It's not only about the `*`, but the `.*`.

Ohh wait, this does not even work with `sphinxcontrib.test_reports.*`, as `.` is a special character. `.` must get escaped: `sphinxcontrib\.test_reports\..*` .

For some examples see: https://regex101.com/r/XTnHX6/1
image

Such complex `target_option` values are not covered by the test cases, as there simpler targets like `b` are used.

tests/test_tr_link.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    Return an empty list when the target option is missing in all items of needs.
+    """
+    assert (
+        tr_link(
+            app=None,
+            need={"id": "1", "a": "1"},
+            needs={"x": {"id": "123"}},
+            test_option="a",
+            target_option="b",
```
danwos, line 23: `target_option` should also contain some more complex regex patterns in some test cases.
For Example:
* `package\.module\.function`
* `package\.module\..*` For all functions of a module
* `package\.module\.(x|y)` Just functions `x` or `y`

danwos (changes requested):
Sorry for the annoying delay in my review.

I just found an important problem in the handling of the regex pattern.
See comments below.

This would also be a breaking change, as `target_option` now gets interpreted as regex pattern, and string-based matches may need an update, if the string contains special regex-chars like `.` or so.

Maybe it would be a good idea to introduce another option or `target_option_regex` to activate regex support by the user and to be backward compatible again.

Looking forward on your thoughts about this :)

However, still thanks a lot for the effort you have already spent on this PR :+1:

cuinixam:
No need to apologise, I totally understand... one cannot handle everything at once and must set priorities 😀
This change is not a blocking point for our projects because I have added the modified `tr_link` method to `conf.py` and everything works as expected. Of course, copying the `tr_link` method in all projects is not a long term solution and therefore I am looking forward to merging this PR 😃

Before I start making any changes, I think we should first define the scope of this PR... like which features will be added for the end user.
The scope initially was to only link one test case to multiple test results by using `.*` in the target option. That is why the regex matching was only enabled if the target string contained a `*`. 

If we want to offer full regex support, it is a completely different story. I would suggest perhaps adding a `tr_link_match` for using regular expressions. As you said, changing 'tr_link' to interpret the 'target_option' as a regex pattern is a breaking change. So if we add an option to globally enable regex for the target option, it will imply that existing 'tr_link' may yield a different result.

danwos:
You are right, The PR gets more complex with every comment. Sorry for that, but I think this is needed to support every aspect a user may see in an option supporting regex or `*`.

However, keeping `tr_link` as it is and providing another option like `tr_link_match` is a good idea.
I would call it `tr_link_regex` to make clear regex is fully supported here.

If `tr_link_regex` is set, `tr_link` is not taken into account anymore.
This should be completely backward compatible.

cuinixam:
By a `tr_link_regex` option you mean adding a bool configuration option and if enabled, `tr_link` will interpret the `target_option` as a regex pattern? Or do you mean that there should be another method `tr_link_regex` that the user can use instead of `tr_link`?

I think I would prefer to add the bool option. The job of `tr_link` is to link "needs" and you could now configure it to use regular expressions.

ubmarco:
@cuinixam Can you please resolve the conflicts? I made some structural changes and want to see whether the PR tests green. There has not been any activity since more than a year - is this topic still relevant?

cuinixam:
Hi @ubmarco, after your changes it makes more sense to have a separate method to use for matching needs.

There was no activity on the PR because I am using an workaround in our projects (see [avengineers/SPLed/conf.py](https://github.com/avengineers/SPLed/blob/2056205e95497835f8087ae372d4c7e1bbf722a9/conf.py#L83) and I was waiting for an answer regarding on how to continue 😁 

I prefer the name `tr_link_match` but I have no problem renaming it to `tr_link_regex`.

chrisjsewell:
Thank you for the patience on this one. sphinx-test-reports has moved into the Sphinx-Needs repository (useblocks/sphinx-needs#1931), and this pull request is one of the three that were open at the move; useblocks/sphinx-needs#1952 holds the list and the decisions. The verdict there is to carry it across: the move recipe in #1931's description was run on this exact branch and applies with no conflicts and your authorship preserved, so nobody needs to redo the work. One design question is open first, whether the feature lands as the `tr_link_regex` flag this branch adds or as a separate `tr_link_match` function, and the maintainer will settle it before the move so you are not asked to change it twice.

Issue #69, opened 2024-01-12
Deprecation warning


kreuzberger:
Minor warning issue 

`
\sphinxcontrib\test_reports\environment.py:10: RemovedInSphinx80Warning: The alias 'sphinx.util.status_iterator' is deprecated, use 'sphinx.util.display.status_iterator' instead. Check CHANGES for Sphinx API modifications.
  from sphinx.util import status_iterator  # NOQA Sphinx 1.5
`

PR #70, opened 2024-01-12
[#69] fix deprecation warning

kreuzberger:
Sphinx in  Versions >= 7 migrated to PathLib and generates warnings.
Tests must considers this too

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,7 +2,7 @@ import nox
 from nox import session
 
 PYTHON_VERSIONS = ["3.8", "3.9", "3.10"]
-SPHINX_VERSIONS = ["5.0"]
+SPHINX_VERSIONS = ["5.0", "6.2.1", "7.2.5"]
 TEST_DEPENDENCIES = [
     "pytest",
     "pytest-xdist",
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,7 +2,7 @@ import nox
 from nox import session
 
 PYTHON_VERSIONS = ["3.8", "3.9", "3.10"]
-SPHINX_VERSIONS = ["5.0", "6.2.1", "7.2.5"]
+SPHINX_VERSIONS = ["5.0", "6.2.1", "7.1.2", "7.2.5"]
 TEST_DEPENDENCIES = [
     "pytest",
     "pytest-xdist",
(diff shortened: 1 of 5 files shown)


noxfile.py
```diff
@@ -2,7 +2,7 @@
 from nox import session
 
 PYTHON_VERSIONS = ["3.8", "3.9", "3.10"]
-SPHINX_VERSIONS = ["5.0"]
+SPHINX_VERSIONS = ["5.0", "6.2.1", "7.1.2", "7.2.5"]
```
danwos, line 5: For Sphinx-Needs we have this matrix settings:
image

Should we use the same versions here as well?

Are there any important changes between `7.1.2` and `7.2.5` which will let our tests behave differently?

tests/conftest.py
```diff
@@ -1,30 +1,40 @@
...
+from pathlib import Path
 from tempfile import mkdtemp
 
 import pytest
-from sphinx.testing.path import path
+from packaging.version import Version
+from sphinx import __version__ as sphinx_version
+
+if Version(sphinx_version) < Version("7.2"):
+    from sphinx.testing.path import path
```
danwos, line 11: Is this sphinx-related import still needed, if everything got converted to use `Path` from `pathlib`?

danwos (commented):
Looks good. Thanks for the PR :+1: 

Just 1-2 questions :)

noxfile.py
kreuzberger: First of all: i would be happy if all useblocks  provided sphinx extensions could match the same build/testing environment, this would make integration into my used building/testing environment

I choosed 7.1.2 cause this was the last version before the deprecation (introduced with 7.2.0)

It seems to me that with 7.2.5 the App object required by the spinx tests api changed the interface from PathLike Objects to pathlib.Paths. So testing one version AFTER this change and one before seems useful.

tests/conftest.py
kreuzberger: i think so cause the same test is made in the conversion to decide if a pathlib.Path object or a sphinx.testing.path object should be returned. 

If you find a more pythonic solution, glad to see and learn. I am still 95% c++/qt developer :smile:

cpolzer:
HEy hello, is there any chance we get this merged?

PR #71, opened 2024-01-19
upgrade to sphinx 7.2.6 and sphinx 6.2.1

kreuzberger:
Due to changes in sphinx 7.2 all conftest.py files should be adapted to accept pathlib.Path objects


kreuzberger:
Ups. not noticed there was another pull request open to fix this :-1: 
https://github.com/useblocks/sphinx-test-reports/pull/70
I think this one is a better solution than the last

kreuzberger:
integrate solution int #70

PR #72, opened 2024-01-19
 add junit ctest xml tests

kreuzberger:
Add test for junit parser in ctest junit xml file for skipped tests
in ctests disabled state is not handled in junit file. Could be matched to skipped by junit parser, decided to skip tests in cmake itself instead of disable the tests


danwos:
Thanks for the PR. :+1: 
Looks good.

PR #76, opened 2024-06-12
add missing dependencies

jsqu4re:
Currently for tests no dependencies are declared. Additionally `plantuml` is used but not listed as a dependency.


danwos:
Sorry for the really late response.
I may miss something, but plantuml is not needed for Sphinx-Test-Reports itself, just for the docs and some tests.
So I would not add it as tool dependency.

PR #83, opened 2024-12-06
Changed import to enable sphinx-needs 4.1 compatibility

MaximilianSoerenPollak:
This is the third 'split' PR that was discussed for splitting the big one into smaller ones. This one here https://github.com/useblocks/sphinx-test-reports/pull/80

This PR changes the import of `make_hashed_id`  to the now only internal version `_make_hashed_id` and adapts the arguments to fit as well.
Unsure if this is the best way to do it, or if maybe upstream the function should be exported again.

fixes #78 

---

Note: This PR has a similar issue as #82 issue being that the 8.1.3 Sphinx tests fail, with the same error message. 
```
FAILED tests/test_custom_template.py::test_custom_koi8_template[test_app0] - sphinx.errors.ThemeError: An error happened in rendering the page index.
FAILED tests/test_custom_template.py::test_custom_utf8_template[test_app0] - sphinx.errors.ThemeError: An error happened in rendering the page index.
FAILED tests/test_custom_template.py::test_custom_template[test_app0] - sphinx.errors.ThemeError: An error happened in rendering the page index.
```
More exact:
```
E           sphinx.errors.ThemeError: An error happened in rendering the page index.
E           Reason: TemplateNotFound("'about.html' not found in ['/tmp/tmptxv71yyp/custom_tr_template/_templates', '/home/maxi/tmp/testing/sphinx-test-reports/.nox/tests-3-12-sphinx-8-1-3/lib/python3.12/site-packages/sphinx/themes/basic']")

.nox/tests-3-12-sphinx-8-1-3/lib/python3.12/site-packages/sphinx/builders/html/__init__.py:1201: ThemeError
```

also here let me know if this can be resolved somehow, as I did not find the 'html' it talks about. 

I could remove `about.html` from the conf.py but unsure if that has further consequences.

PR #80, opened 2024-11-26
Enabling Sphinx-needs 4.1 compatibility & simplifying setup

MaximilianSoerenPollak:
- Upgraded to allow for compaitiblity with Sphinx-needs 4.0+
- Changed paths  to allow for better integration with other tools (like bazel).
- Added setuptools as dependency as it wasn't there already but is used.
- Added possibility for mutliple values in 'tr_link' targeted option
- Enabled TestFileDirective, TestSuiteDirective, TestCaseDirective to
  have additional options from extra_links & extra_options
- Edited tests to only test newer versions of Sphinx & Python
- Fixed tests to account for new paths

Addresses: #78 #79 

---
### Please note: 
Some of the tests do not pass the `about.html` is the problem, but all functionality seems to be right and present. 
I'm sure there is better ways to do some of the things that makes more sense, this should be seen as 'proof of concept' and starting point. 
There also is an error I could not figure out: 
```
sphinxcontrib/test_reports/environment.py:70: RemovedInSphinx90Warning: 'sphinx.builders.html.StandaloneHTMLBuilder.css_files' is deprecated. Check CHANGES for Sphinx API modifications.
```
Not quite sure where that comes from, or how do fix this one. It happens with all of the environment functions, not just this one

PR #84, opened 2024-12-06
Added possibility for directives to have 'extra-options' defined in c…

MaximilianSoerenPollak:
This is the fourth 'split' PR that was discussed for splitting the big one into smaller ones. This one here https://github.com/useblocks/sphinx-test-reports/pull/80

Note: PR depends on another. #83. 

This PR adds the capability of the directives declared in the package to contain `extra_options` defined via the `sphinx-needs` option in conf.py. 

Note: This PR has a similar issue as #83  issue being that the 8.1.3 Sphinx tests fail, with the same error message.
```
FAILED tests/test_custom_template.py::test_custom_koi8_template[test_app0] - sphinx.errors.ThemeError: An error happened in rendering the page index.
FAILED tests/test_custom_template.py::test_custom_utf8_template[test_app0] - sphinx.errors.ThemeError: An error happened in rendering the page index.
FAILED tests/test_custom_template.py::test_custom_template[test_app0] - sphinx.errors.ThemeError: An error happened in rendering the page index.
```
More exact:
```
E           sphinx.errors.ThemeError: An error happened in rendering the page index.
E           Reason: TemplateNotFound("'about.html' not found in ['/tmp/tmptxv71yyp/custom_tr_template/_templates', '/home/maxi/tmp/testing/sphinx-test-reports/.nox/tests-3-12-sphinx-8-1-3/lib/python3.12/site-packages/sphinx/themes/basic']")

.nox/tests-3-12-sphinx-8-1-3/lib/python3.12/site-packages/sphinx/builders/html/__init__.py:1201: ThemeError
```

### DIFF
--- a/sphinxcontrib/test_reports/directives/test_case.py
+++ b/sphinxcontrib/test_reports/directives/test_case.py
@@ -2,6 +2,7 @@ from docutils import nodes
 from docutils.parsers.rst import directives
 from sphinx_needs.api import add_need
 from sphinx_needs.utils import add_doc
+from sphinx_needs.config import NeedsSphinxConfig
 
 from sphinxcontrib.test_reports.directives.test_common import TestCommonDirective
 from sphinxcontrib.test_reports.exceptions import TestReportInvalidOption
@@ -32,9 +33,6 @@ class TestCaseDirective(TestCommonDirective):
     }
 
     final_argument_whitespace = True
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
 
     def run(self, nested=False, suite_count=-1, case_count=-1):
         self.prepare_basic_options()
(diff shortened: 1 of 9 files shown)


danwos:
I'm not sure, if I like this change because of its high connection to Sphinx-Needs and its config usage.
It may force us to react on new releases of Sphinx-Needs, if config or some options-lists get extended.

Also allowing all options without any checks can bring some trouble or even be in conflict.

I see two additional approaches:
1) Make use of [needextend](https://sphinx-needs.readthedocs.io/en/latest/directives/needextend.html) to manipulate created test objects and add the needed data.
2) Have a config option like `tr_extra_options` to define selected options. Only these can then be used.

MaximilianSoerenPollak:
You are right there, if we would just do it as I have proposed it probably would tie too closely to sphinx-needs and need changes sooner. 

For the first proposal, would it then be that you do `needextend` inside the RST file where you use the test objects, if I understand correctly? 

And the 'tr_extra_options' would then be an additional configuration value inside the Extension that the user can set, and we would add this to the node / classes to make sure those are 'allowed' and don't error? 
That approach to me seems the 'nicest' as it allows the user to just have one line of code instead of having to add 'needextend' to all of the files containing test objects. 

Which one would you prefer?

danwos:
For me, `tr_extra_options` is also the best approach,  so `needextend` can be avoided.
It may be some extra configuration for the user, but in the end:

* the users can handle naming conflicts on their own
* the code is not using Sphinx-Needs internal stuff, which eventually will end up in Sphinx-Needs-version specific implementations.

MaximilianSoerenPollak:
So we would have an option `tr_extra_options`  that all the directives are allowed to use that the user can define. 
A question regarding that, should this 'internally' register these options as extra options as well? Or should we just make a note to the user that these have to be registered by them themselves?

MaximilianSoerenPollak:
Can be closed, as new PR will be made for this. 
See #96

PR #91, opened 2025-01-17
using hatch instead of setuptools

danwos:


danwos:
Does currently not work, as Sphinx can't find the needed `setup()` function of `sphinxcontrib.spehinx_test_reports`.

ubmarco:
Closed in favor of https://github.com/useblocks/sphinx-test-reports/pull/110

PR #99, opened 2025-03-26
Allows JSON Parser to set additional options

danwos:
The JSON Parser mapping configuration supports now a mapping between any configured Sphinx-Needs options and a JSON value.

This allows to set options like id, status, tags and extra_options with data from a JSON file.


docs/configuration.rst
```diff
@@ -187,8 +187,40 @@ Given the above JSON example, the following "selector" will address the value ``
 
    ["level_1", "level_2", 1, "value"]
 
+Additional meta-data
+~~~~~~~~~~~~~~~~~~~~
+The JSON Parser allows to set additional options for the created test-case, for instance
+a status, tags or even the ID of the test case.
 
-**Example config**
+Just use the name in the ``tr_json_mapping`` and make sure none standard options
```
ubmarco, line 195: Unclear wording. Do you mean this? If yes, why do users have to make sure?
```suggestion
Just use the name in the ``tr_json_mapping`` and make sure no standard options
```

doc-requirements.txt
```diff
@@ -4,4 +4,5 @@ sphinx-test-reports>=0.3.3
 sphinxcontrib-plantuml
 sphinx-design
 sphinx-immaterial
-pillow
\ No newline at end of file
+pillow
+pre-commit
```
ubmarco, line 8: should go to a dedicated PR - why would pre-commit be a dependency of docs build?

sphinxcontrib/test_reports/directives/test_case.py
```diff
@@ -148,6 +149,18 @@ def run(self, nested=False, suite_count=-1, case_count=-1):
         if case_parameter is None:
             case_parameter = ""
 
+        # Set extra data, which is not part of the Sphinx-Test-Reports default options
+        for key, value in case.items():
```
ubmarco, line 153: does any code part do input validation of the input json, e.g. is there a schema that checks the value is actually a string?

ubmarco (changes requested):
Just went through the code without fully understanding all details :/

doc-requirements.txt
danwos: removed

ubmarco (approved):

PR #110, opened 2025-04-01
🔧 Use flit

ubmarco:
Migrate to flit / uv


danwos (approved):
Nice, lets give `flit` a try. Never used it, but read about it several times.

PR #115, opened 2025-04-01
🐛 Fix plantuml on RTD

ubmarco:
Enable fail_on_warning


danwos (approved):
Thanks for the fix :rocket:

PR #133, opened 2026-03-23
🔧 Add sphinx-needs 8.0.0 compatibility

cpolzer:
## Summary

- Replace deprecated `add_extra_option()` with `add_field()` for sphinx-needs >= 8.0.0, using a version-gated shim that falls back to `add_extra_option()` for older versions
- Add sphinx-needs `8.0.0` to the CI matrix and nox test matrix
- Fix warning checks in tests to inspect `stderr` (where Sphinx actually writes warnings, not `stdout`)
- Add `test_test_file_renders_need_with_counts` to catch regressions where int kwargs to `add_need()` are rejected

## Test plan

- [ ] Run `nox -s "tests-3.12(sphinx='8.1.3', sphinx_needs='8.0.0')"` and verify all tests pass
- [ ] Run against an older sphinx-needs version (e.g. `6.3.0`) to confirm backwards compatibility

## Reason:
Upgrading another sphinx-needs project showed errors:

"sphinx-test-reports passes integer field values (e.g. suites=1) which sphinx-needs 8.0 rejects with InvalidNeedException. "

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -107,7 +107,7 @@ introduces several maintenance improvements.
 -----
 :Released: 26.09.2022
 
-* Improvement: Supporting `Sphinx-Needs <https://www.sphinx-needs.com/>`__ ``>= 1.01`` only.
+* Improvement: Supporting `Sphinx-Needs <https://sphinx-needs.readthedocs.io/en/latest/>`__ ``>= 1.01`` only.
 * Improvement: Migrated nosetests to pytest.
 
 0.3.7
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,6 +17,9 @@ import os
 # import os
 import sys
 
+import sphinx_needs
+from packaging.version import Version
+
 sys.path.append(os.path.abspath("."))
 
 from ub_theme.conf import html_theme_options
@@ -67,7 +70,13 @@ plantuml_output_format = "png"
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates", "ub_theme/templates"]
-needs_extra_options = ["more_info"]
+if Version(sphinx_needs.__version__) >= Version("8.0.0"):
+    needs_fields = {
+        "more_info": {"nullable": True},
+    }
+else:
+    needs_extra_options = ["more_info"]
+
 tr_extra_options = ["more_info"]
 # Add a custom test report template. Please add a relative path from this conf.py
 # tr_report_template = "./custom_test_report_template.txt"
--- a/sphinxcontrib/test_reports/test_reports.py
+++ b/sphinxcontrib/test_reports/test_reports.py
@@ -191,14 +191,13 @@ def sphinx_needs_update(app: Sphinx, config: Config) -> None:
     needs_version = Version(sphinx_needs.__version__)
     use_schema = needs_version >= Version("6.0.0")
 
-    _register_field(app, getattr(config, "tr_file_option", "file"))
-    _register_field(app, "suite")
-    _register_field(app, "case")
-    _register_field(app, "case_name")
-    _register_field(app, "case_parameter")
-    _register_field(app, "classname")
-
     if use_schema:
+        _register_field(app, getattr(config, "tr_file_option", "file"), schema={"type": "string"})
+        _register_field(app, "suite", schema={"type": "string"})
+        _register_field(app, "case", schema={"type": "string"})
+        _register_field(app, "case_name", schema={"type": "string"})
+        _register_field(app, "case_parameter", schema={"type": "string"})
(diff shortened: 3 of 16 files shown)


sphinxcontrib/test_reports/test_reports.py
```diff
@@ -9,7 +9,18 @@
...
 
 # from docutils import nodes
-from sphinx_needs.api import add_dynamic_function, add_extra_option, add_need_type
+from sphinx_needs.api import add_dynamic_function, add_need_type
+
+try:
+    from sphinx_needs.api import add_field as _add_field
+
+    def _register_field(app, name, schema=None):
+        _add_field(name, name, schema=schema)
```
patdhlk: The `add_field()` signature is `add_field(name, /, description, *, schema=...)`. Passing `name` as the description works but is semantically wrong — all field descriptions will just be the raw field name.

docs/conf.py
```diff
@@ -17,6 +17,9 @@
 # import os
 import sys
 
+import sphinx_needs
+from packaging.version import Version
```
patdhlk, line 21: `packaging` is not a declared dependency, but used in all (also test) `conf.py` files now. 

`from packaging.version import Version` is used unconditionally in `test_reports.py` and now in 8 `conf.py` files, but `packaging` isn't in `pyproject.toml` dependencies. It works via transitive deps from Sphinx/pip, but is fragile.

tests/test_test_file.py
```diff
@@ -36,5 +36,26 @@ def test_test_file_needs_extra_options_no_warning(test_app):
...
+
+    Catches regressions where int values (suites, cases, passed, etc.) passed
+    to add_need() are rejected by newer sphinx-needs versions.
+    """
+    app = test_app
+    app.build()
+    assert app.statuscode == 0
+
+    html = Path(app.outdir, "index.html").read_text()
+    assert "TESTFILE_1" in html
```
patdhlk, line 68: `test_test_file_renders_need_with_counts` only checks `TESTFILE_1` in `html` - doesn't verify integer count fields were actually rendered, despite the docstring claiming it catches integer rejection regressions

tests/test_test_file.py
```diff
@@ -36,5 +36,26 @@ def test_test_file_needs_extra_options_no_warning(test_app):
 
     assert out.returncode == 0
 
-    # Check no warnings
+    # Check no warnings — Sphinx writes warnings to stderr, not stdout
     assert "WARNING" not in out.stdout.decode("utf-8")
+    assert "WARNING" not in out.stderr.decode("utf-8")
+    assert "ERROR" not in out.stderr.decode("utf-8")
```
patdhlk: Test still asserts `WARNING not in stdout` after noting Sphinx writes warnings to stderr

patdhlk (changes requested):
Thanks for the PR, some small nitpicks. 

One general thing I noted: same version-gating pattern copy-pasted across 7 test conf.py files. A shared helper would reduce future maintenance. But this can be picked up in the future

sphinxcontrib/test_reports/test_reports.py
cpolzer: Hey @patdhlk,  enhanced with a better description mapping

docs/conf.py
cpolzer: `packaging` is now part of the dependency tree implicitly

tests/test_test_file.py
cpolzer: tests updated/extended

tests/test_test_file.py
cpolzer: fixed

patdhlk:
@cpolzer only one ci issue then we are ready to go

docs/changelog.rst
```diff
@@ -107,7 +107,7 @@ introduces several maintenance improvements.
 -----
 :Released: 26.09.2022
 
-* Improvement: Supporting `Sphinx-Needs <https://www.sphinx-needs.com/>`__ ``>= 1.01`` only.
+* Improvement: Supporting `Sphinx-Needs <https://sphinx-needs.readthedocs.io/en/latest/>`__ ``>= 1.01`` only.
```
cpolzer, line 110: @patdhlk i changed it, but looking at the error in ci - was https://www.sphinx-needs.com/ ever available and was shut down? or is hosting broken?

The almighty dumpster seems to know that domain: 

Image

cpolzer: the error was/ is a cert chain error...

patdhlk: I remember it being available, but it redirects to https://useblocks.com/open-source/sphinx-needs now which might have caused the cert chain error 🤔

patdhlk (approved):

PR #135, opened 2026-05-12
feat: link JUnit XML properties to sphinx-needs fields

cpolzer:
## Summary

This PR adds support for JUnit XML `<properties>` elements, enabling **requirement traceability** by mapping test case and test suite properties to sphinx-needs fields.

The entire properties feature is opt-in. If you don't set tr_extra_options or tr_property_link_types, the extension behaves exactly as it did before this PR.

### Use Case

Many test frameworks (e.g., pytest with custom markers, Maven Surefire, JUnit 5) can embed key-value properties in JUnit XML output. These properties often contain requirement IDs, priorities, or categories. Previously, `sphinx-test-reports` ignored this data. Now you can surface it directly in your documentation and link tests back to requirements.

### Example JUnit XML

```xml
<testcase classname="auth.tests" name="test_login" time="0.42">
  <properties>
    <property name="verifies" value="REQ_AUTH_001,REQ_AUTH_002"/>
    <property name="priority" value="high"/>
  </properties>
</testcase>
```

### New Configuration Options

| Option | Type | Description |
|--------|------|-------------|
| `tr_extra_options` | `list[str]` | Which JUnit property names become sphinx-needs fields |
| `tr_property_link_types` | `dict[str, str]` | Maps a property name to a sphinx-needs link field (e.g. `"links"`) |

### Example `conf.py`

```python
# Make these JUnit properties available as need fields
tr_extra_options = ["verifies", "priority", "category"]

# Map the "verifies" property to sphinx-needs links
tr_property_link_types = {
    "verifies": "links",  # comma-separated IDs → need links
}
```

### What Changed

- **Parser** (`junitparser.py`): Extracts `<properties>` from both `<testcase>` and `<testsuite>` elements with a safe guard against empty `<properties/>`.
- **Test Case Directive** (`directives/test_case.py`): Flattens allowed properties into top-level fields; processes `tr_property_link_types` to build semicolon-separated sphinx-needs link values.
- **Test Suite Directive** (`directives/test_suite.py`): Same property flattening and link mapping at the suite level.
- **Field Registration** (`test_reports.py`): Automatically registers `tr_extra_options` as sphinx-needs fields on startup. Duplicate registration is handled gracefully (logs debug, no error).
- **Config** (`config.py`): Added `"properties"` to the default JUnit parser options list.
- **Tests** (`test_properties.py`): 384 lines of tests covering parser extraction, empty properties, integration build, DOM rendering, link creation, and backward compatibility.

### Backward Compatibility

- Fully backward compatible. If `tr_extra_options` and `tr_property_link_types` are not set, behavior is identical to before.
- Property flattening is filtered through the `tr_extra_options` allowlist, so unknown properties never leak into `add_need()`.

### Test Coverage

- Parser unit tests for property extraction
- Empty `<properties/>` handling
- Integration build with real JUnit XML
- HTML DOM assertions for rendered fields
- Link creation and merging
- Backward compatibility (no config → no change)

### DIFF
--- a/sphinxcontrib/test_reports/directives/test_case.py
+++ b/sphinxcontrib/test_reports/directives/test_case.py
@@ -160,6 +160,39 @@ class TestCaseDirective(TestCommonDirective):
         if case_parameter is None:
             case_parameter = ""
 
+        # Flatten JUnit <properties> into top-level case keys so that
+        # the extra-data loop below picks them up as sphinx-needs fields.
+        # Only propagate properties that are explicitly listed in
+        # tr_extra_options to avoid unknown-kwarg errors from add_need.
+        # Properties do not overwrite core JUnit attributes (name, time, etc.).
+        allowed_extras = set(getattr(self.app.config, "tr_extra_options", []))
+        case_properties = case.get("properties", {})
+        for prop_name, prop_value in case_properties.items():
+            if prop_name in allowed_extras and prop_name not in case:
+                case[prop_name] = prop_value
+
+        # Process tr_property_link_types: map property values to sphinx-needs links
+        tr_property_link_types = getattr(self.app.config, "tr_property_link_types", {})
+        for prop_name, link_field in tr_property_link_types.items():
+            prop_value = case_properties.get(prop_name, "")
+            if prop_value:
+                # Convert comma-separated IDs to semicolon-separated (sphinx-needs link format)
+                link_ids = ";".join(
+                    id_val.strip() for id_val in prop_value.split(",") if id_val.strip()
+                )
+                if link_field == "links":
+                    existing = self.test_links
+                else:
+                    existing = self.extra_options.get(link_field, "")
+                if existing:
+                    merged = existing + ";" + link_ids
+                else:
+                    merged = link_ids
+                if link_field == "links":
+                    self.test_links = merged
+                else:
+                    self.extra_options[link_field] = merged
+
         # Set extra data, which is not part of the Sphinx-Test-Reports default options
         for key, value in case.items():
             if key == "id" and value not in ["", None]:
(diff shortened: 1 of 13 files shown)


sphinxcontrib/test_reports/directives/test_case.py
```diff
@@ -160,6 +160,39 @@ def run(self, nested=False, suite_count=-1, case_count=-1):
...
+                else:
+                    existing = self.extra_options.get(link_field, "")
+                if existing:
+                    merged = existing + ";" + link_ids
+                else:
+                    merged = link_ids
+                if link_field == "links":
+                    self.test_links = merged
+                else:
+                    self.extra_options[link_field] = merged
```
patdhlk: @cpolzer same code/logic as in `test_suite.py` L76-L110

patdhlk (changes requested):
Thanks for the PR. In addition to the duplication of the file we need some documentation. I see no changes to docs/ or README. The user-facing PR description is good, but those config options should be in the rendered docs.

sphinxcontrib/test_reports/directives/test_case.py
cpolzer: @patdhlk moved to "test_common .py" to deduplicate.

cpolzer:
> Thanks for the PR. In addition to the duplication of the file we need some documentation. I see no changes to docs/ or README. The user-facing PR description is good, but those config options should be in the rendered docs.

@patdhlk added user facing docs - sphinx ,  skipped readme it has no configuration content as of now.

patdhlk (approved):
LGTM 🚀

StephenWithPH:
This is _exactly_ what I need. Thank you for the PR.

I think there is a slight error:

https://github.com/useblocks/sphinx-test-reports/blob/703197241ec10408344b2fb15ac3c7d2e5300deb/docs/configuration.rst#L167-L169

The current deployed docs show this as available since 1.3.0; 1.3.2 (current on PyPI) was released 11/13/25.

@patdhlk ... possible to cut a release to get this shipped?

PR #136, opened 2026-05-25
Clarify Sphinx-Needs type names in docs

### LINKED ISSUE #131: Possible inconsistencies in documentation
Hi, I'm new to this package (and sphinx-needs in general), so forgive me if I'm missing anything obvious.

While trying to display test cases and suites from a xml file generated with pytest following the [examples in your documentation](https://sphinx-test-reports.readthedocs.io/en/stable/filter.html#needtable-a-table-for-all-tests), I came across the `needtable` directive. The example use `:types: test-file` and also mentions `test-suite` and `test-case` types. In my case, I needed the latter two. However, the rendered page would always show "No needs passed the filters" (as does the example on your documentation!). When omitting the `:types:` entirely, I'd indeed get a table with both the test cases and test suite I expected to see. In the `type` column, it would show "test-case" and "test-suite", respectively, matching the documentation. After some digging in the raw html however, I discovered that it actually had "testcase" and "testsuite" (no hyphen) in the table! The hyphenated versions seem to originate from automatic line breaks in the rendered html. Using `:types: testcase` now indeed produced the expected table.

I don't know which way around it is meant to be, but I think the documentation should match the actual behavior, either way around. Then I'd also expect the various examples shown in https://sphinx-test-reports.readthedocs.io/en/stable/filter.html to no longer be empty.

Additionally, I discovered that the [docs page about the `test-case` directive](https://sphinx-test-reports.readthedocs.io/en/stable/directives/test_case.html) reads:

> `test-suite` creates a need of type `test-suite` and adds the following options automatically:

I'm assuming this is a simple copy-past mistake, but confused me for a second when I first encountered it.

Furthermore, the `pytest` example on the [docs page about the `test-results` directive](https://sphinx-test-reports.readthedocs.io/en/stable/directives/test_results.html) currently shows only skipped tests. Is that intentional?

Mirochill:
## Summary
- update filtering examples to use the Sphinx-Needs type names `testfile`, `testsuite`, and `testcase`
- add a note distinguishing those type names from the hyphenated directives `test-file`, `test-suite`, and `test-case`
- fix the `test-case` directive page copy/paste text so it describes `test-case` and the `testcase` type
- update the `needlist` examples to compare the numeric `cases` field directly

Fixes #131

## Validation
- `git diff --check -- docs/filter.rst docs/directives/test_file.rst docs/directives/test_suite.rst docs/directives/test_case.rst`
- Read the Docs build succeeded: https://sphinx-test-reports--136.org.readthedocs.build/en/136/
- Not run locally: tests (documentation-only change)

### DIFF
--- a/docs/directives/test_case.rst
+++ b/docs/directives/test_case.rst
@@ -10,7 +10,7 @@ Usage
 
 .. code-block:: rst
 
-   .. test-case:: My Test Suite
+   .. test-case:: My Test Case
       :file: my_test_data.xml
       :suite: my_tested_suite
       :case: my_case
@@ -36,7 +36,7 @@ to specify only ``case`` or ``classname``. It depends on the loaded test-data, i
 or if it selects only the first found test case. The best case is to always try to specify both values, ``case`` and
 ``classname``.
 
-``test-suite`` creates a need of type ``test-suite`` and adds the following options automatically:
+``test-case`` creates a need of type ``testcase`` and adds the following options automatically:
 
 * **result**: Result of the test case run. E.g passed or failed.
 * **time**: Needed time for running the test case
--- a/docs/directives/test_file.rst
+++ b/docs/directives/test_file.rst
@@ -28,7 +28,7 @@ The following options can be set:
   Must be used together with ``auto_suites``!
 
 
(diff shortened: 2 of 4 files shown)


ubmarco (approved):
LGTM, thanks

PR #141, opened 2026-08-31
feat: support the googletest XML dialect

ubmarco:
First of a four-PR stack that makes `sphinx-test-reports` able to replace a hand-rolled test-link extension. Each PR is independently reviewable; this one is parser-only.

googletest deviates from the pytest/nose JUnit flavour in ways that silently produced wrong or incomplete data:

- **`status="notrun"` read as `passed`.** A disabled googletest case has no result child element, so it fell through the `skipped`/`failure`/`error` checks into `passed`. It is now `disabled`; the googletest `result="suppressed"` attribute is accepted as an equivalent signal.
- **Only the first failure was kept.** googletest emits one `<failure>` per failed assertion, and `GTEST_SKIP` can fire repeatedly. All parts are now collected under a new `parts` key, each with kind, type, message and body text.
- **`RecordProperty` in attribute form was invisible.** googletest wrote test-case properties as attributes before 1.8.1 (the form its official docs still show), and suite-level properties stayed attribute-only up to 1.15.x. Unknown attributes on `<testcase>`/`<testsuite>` are now read as properties; `<properties>` children remain the modern form and win on conflict.
- **`<system-err>` was dropped entirely**, and `<system-out>` read only its first block. Both are now collected and joined.
- `timestamp`, `value_param` and `type_param` are parsed.

## Backwards compatibility

The flat `type`/`text`/`message` keys still report the first part, including their historical per-kind defaults (`message` defaults to `""` for a failure and `"unknown"` for skipped/error), so existing consumers see no change. The result vocabulary keeps `failure` — it is a documented need field value and the `tr_failure` CSS class — and only gains `disabled`.

Every new key defaults to `""` rather than being absent, so a consumer can emit each field unconditionally, and all of them are added to `DEFAULT_OPTIONS`; otherwise a new parsed key would reach `add_need` as an unexpected keyword argument.

## Verification

10 new tests against a googletest fixture covering the attribute matrix, plus the existing 61 — 71 passing, mypy strict and pre-commit clean, docs build clean with `-W`.

### DIFF
--- a/sphinxcontrib/test_reports/junitparser.py
+++ b/sphinxcontrib/test_reports/junitparser.py
@@ -50,13 +50,15 @@ TESTSUITE_KNOWN_ATTRIBUTES = frozenset(
 RESULT_PART_KINDS = ("skipped", "failure", "error")
 
 
-def _child_tag(child):
+def _child_tag(child: objectify.ObjectifiedElement) -> str | None:
     """Tag name of an lxml child, or ``None`` for comments and instructions."""
     tag = child.tag
     return tag if isinstance(tag, str) else None
 
 
-def _collect_properties(xml_object, known_attributes):
+def _collect_properties(
+    xml_object: objectify.ObjectifiedElement, known_attributes: frozenset[str]
+) -> dict[str, str]:
     """Read ``RecordProperty`` values from both dialect forms.
 
     Unknown attributes are the legacy/attribute form; ``<properties>`` children
@@ -77,7 +79,9 @@ def _collect_properties(xml_object, known_attributes):
     return properties
 
 
-def _collect_result_parts(testcase):
+def _collect_result_parts(
+    testcase: objectify.ObjectifiedElement,
+) -> list[dict[str, str]]:
     """Every ``<failure>``/``<error>``/``<skipped>`` child, in document order.
 
     googletest emits one element per failed assertion and ``GTEST_SKIP`` can
@@ -85,7 +89,7 @@ def _collect_result_parts(testcase):
     The per-part ``type``/``message`` defaults match what the flat, historical
     ``type``/``message`` keys have always reported for that kind.
     """
-    parts = []
+    parts: list[dict[str, str]] = []
     for child in testcase.iterchildren():
         kind = _child_tag(child)
         if kind not in RESULT_PART_KINDS:
@@ -103,7 +107,7 @@ def _collect_result_parts(testcase):
     return parts
 
 
-def _collect_captured_output(xml_object, tag):
+def _collect_captured_output(xml_object: objectify.ObjectifiedElement, tag: str) -> str:
     """Join every ``<system-out>``/``<system-err>`` block of an element."""
     blocks = [
         child.text or ""
@@ -189,9 +193,10 @@ class JUnitParser:
                 tc_dict["text"] = first_part["text"]
                 tc_dict["message"] = first_part["message"]
             else:
+                # No result child at all, so the case either passed or never
+                # ran; failures, errors and skips are all handled above.
                 # googletest reports a disabled test as status="notrun"
-                # (result="suppressed") with no result child at all -- which
-                # otherwise reads as a pass.
+                # (result="suppressed"), which otherwise reads as a pass.
                 disabled = (
                     testcase.attrib.get("status") == "notrun"
                     or testcase.attrib.get("result") == "suppressed"
--- a/sphinxcontrib/test_reports/junitparser.py
+++ b/sphinxcontrib/test_reports/junitparser.py
@@ -6,6 +6,112 @@ import os
 
 from lxml import etree, objectify
 
+#: Attributes the JUnit/googletest dialects define themselves. Every *other*
+#: attribute is a ``RecordProperty`` value in attribute form: googletest wrote
+#: test-case properties as attributes before 1.8.1 (the form its official docs
+#: still show), and suite-level properties stayed attribute-only up to 1.15.x.
+TESTCASE_KNOWN_ATTRIBUTES = frozenset(
+    {
+        "assertions",
+        "classname",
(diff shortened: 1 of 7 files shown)


MaximilianSoerenPollak:
One thing that recently came up in Docs-As-Code that I'm not sure if this extensions handles at all, but also not sure if it should be in this PR. 

=> there apprently is a XML result called `error` when the test recieves a `sigterm` or similar signal or crashes during execution. 
This we also should handle if possible, as to not report wrong status.

sphinxcontrib/test_reports/junitparser.py
```diff
@@ -53,53 +159,54 @@ def parse_testcase(xml_object):
...
             else:
-                tc_dict["result"] = "passed"
+                # googletest reports a disabled test as status="notrun"
+                # (result="suppressed") with no result child at all -- which
+                # otherwise reads as a pass.
+                disabled = (
+                    testcase.attrib.get("status") == "notrun"
+                    or testcase.attrib.get("result") == "suppressed"
+                )
+                tc_dict["result"] = "disabled" if disabled else "passed"
```
MaximilianSoerenPollak, line 204: Does this mean for `error` or `failed` we would also put passed, or am I misreading here?

tests/doc_test/utils/gtest_data.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MaximilianSoerenPollak, line 1: failed & error `status` test cases missing, as well as a `passed` testcase without any properties.

Not sure if the `error` should be in the same or new file

sphinxcontrib/test_reports/junitparser.py
```diff
@@ -6,6 +6,112 @@
 
 from lxml import etree, objectify
 
+#: Attributes the JUnit/googletest dialects define themselves. Every *other*
```
MaximilianSoerenPollak, line 9: Unsure what the coding guidelines in this repo / project are, but typehints probably would make some of it easier (especially function returns).

What is your thoughts on this?

sphinxcontrib/test_reports/junitparser.py
```diff
@@ -6,6 +6,112 @@
...
+    """Read ``RecordProperty`` values from both dialect forms.
+
+    Unknown attributes are the legacy/attribute form; ``<properties>`` children
+    are the modern form and therefore win on conflict.
+    """
+    properties = {
+        name: value
+        for name, value in xml_object.attrib.items()
+        if name not in known_attributes
+    }
```
MaximilianSoerenPollak, line 71: If I'm reading this right, when there is `0` names not in the known_attributes this should be an empty dictionary correct?

MaximilianSoerenPollak (changes requested):
Seems like a great addition.

I got some questions to further my understanding and some things are I think maybe missing in the test.xml you have made. 
Though please correct me if that is not needed.

MaximilianSoerenPollak (approved):
For me this looks fine now. 

All issues were addressed.

ubmarco:
Already handled, and this PR improves it.

- `<error>` support landed in 1.4.0 (#134). This PR keeps it and now aggregates **all** `<error>` parts, not just the first.
- Verified against a Bazel-shaped crash report: `result = "error"`, message + body kept, `system-err` captured.
- Framing nit: googletest never emits `<error>` — only `<failure>`. That shape comes from the **runner** (Bazel synthesizes `test.xml` from the log when the binary dies).
- So → new fixture `runner_error_data.xml` (sigterm, exit code, 2-part error), tests in `test_junit_parser.py` next to the existing `<error>` coverage. That answers your "same or new file": new file, different producer.
- Relevant for Docs-As-Code: `score_source_code_linker` does **not** handle `<error>`. `parse_testcase_result()` only checks `skipped`/`failure` and returns `"passed"` otherwise — a sigterm-killed test reports **passed** there today.

Added in 9c452d2.

sphinxcontrib/test_reports/junitparser.py
ubmarco: Misreading — but the truncated hunk invites it.

That `else` only runs when there is **no** `<failure>`/`<error>`/`<skipped>` child at all. Those three are classified in the `if first_part is not None:` branch just above, so they can never reach `"passed"`.

Reworded the comment so the branch says that where you're reading it (df1d7b6).

tests/doc_test/utils/gtest_data.xml
ubmarco: Three separate answers:

- **failed** — already there: `Subtraction`, with two `<failure>` parts.
- **error** — real gap, fixed, but not in this file: googletest never emits `<error>`, the runner does. New fixture `runner_error_data.xml`, tests in `test_junit_parser.py`. → new file, as you suspected.
- **passed without properties** — already there: `Works/0`. Added an explicit assert that its `properties == {}` — it carries `value_param`/`type_param`, which must not leak in as properties.

9c452d2.

sphinxcontrib/test_reports/junitparser.py
ubmarco: Agreed — added for the helpers this PR introduces (df1d7b6).

Caveat: they are not enforced. `junitparser.py` is in mypy's exclude list in `pyproject.toml` (pre-existing TODO). Reactivating the file is its own PR: the lxml objectify boundary is untyped and this repo runs mypy with `disallow_any_expr`, so it needs a typed wrapper first.

sphinxcontrib/test_reports/junitparser.py
ubmarco: Correct — no matching attributes gives `{}`.

Pinned it: `test_a_testcase_without_any_properties_yields_an_empty_dict` (9c452d2).

PR #142, opened 2026-08-31
feat: expose the test-case source location as need fields

ubmarco:
Second of four. **Stacked on #141** — review that one first; this PR's diff against it is small.

The JUnit/googletest `<testcase>` attributes `file` and `line` point at the test source. The parser read them, but they never reached `add_need`.

## The blocker, and the latent bug behind it

The need field `file` already carries the path of the XML *report*. `tr_file_option` exists to rename that field — but it was only half wired: the field *registration* honoured it while all three directives kept passing a hardcoded `file=`. So setting `tr_file_option = "report_file"` did not free the name `file`; it produced a need carrying an unregistered field, and the build failed inside `add_need`. That is fixed here, which is what makes the rest possible.

## What this adds

- All three directives pass the report path under the configured field name, so `tr_file_option` works end to end.
- Two new options name the fields carrying the source location: `tr_source_file_option` (default `case_file`) and `tr_source_line_option` (default `case_line`). Both are registered and always emitted by the `test-case` directive.
- The defaults avoid the collision. A project whose metamodel spells the source location verbatim as `file`/`line` gets there by renaming the report field instead — field names are configuration, not policy.
- The parser's absent-attribute sentinels (`"unknown"`, `-1`) become empty strings, so a report without file/line yields empty fields rather than a fake location. This is the norm for pytest: its default `junit_family = xunit2` filters those attributes off `<testcase>` entirely, and only `xunit1`/`legacy` emit them.
- Two options resolving to the same field name previously produced a bare `TypeError` from inside a directive (`add_need` receiving one keyword twice). It is now an `InvalidConfigurationError` naming both options.

`tr_file_option` was undocumented despite already existing; it is documented here alongside the two new options.

## Verification

11 new tests — two Sphinx fixture projects (default names, and renamed so the source location occupies `file`/`line`) plus unit tests for the collision check. 82 passing, mypy strict and pre-commit clean, docs build clean with `-W`.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,6 +13,12 @@ Unreleased
   in attribute form are read (on ``<testcase>`` for googletest < 1.8.1 and on
   ``<testsuite>`` for suite-level properties up to 1.15.x), and ``timestamp``,
   ``value_param`` and ``type_param`` are parsed. ``<system-err>`` is captured.
+* Feature: The source location of a test case is available as Sphinx-Needs
+  fields, configurable via the new ``tr_source_file_option`` and
+  ``tr_source_line_option``.
+* Bugfix: ``tr_file_option`` is now honoured by the directives, not only by the
+  field registration. Renaming the field previously produced needs carrying an
+  unregistered field.
 
 .. _`release:1.4.0`:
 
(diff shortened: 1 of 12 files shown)


MaximilianSoerenPollak (approved):
Seems fine to me.

PR #143, opened 2026-08-31
feat: opt-in deterministic test-case need IDs

ubmarco:
Third of four. **Stacked on #142.**

## The problem

The default auto-ID hashes *(type, title, content)*, and a test case's content carries its failure text. So the need ID of a test case **changes when the test starts failing differently**. Anything holding a reference breaks: build caches, a link authored from a test specification to its expected result, and every diff of an exported needs.json.

## The scheme

A new Sphinx-free module derives the ID from where the test *is*:

```
{type}__{Classname__Casename}_{shorthash(file + name)}
```

with five lowercase base32 letters of sha256. Sphinx-free on purpose — the same identity has to be derivable from a converter CLI and a build action, not only from the directives.

Enabled with `tr_deterministic_case_ids = True`; **off by default**, because turning it on changes IDs. Both ID-producing paths honour it — the `test-case` directive's own fallback and the parent-scoped ID that `:auto_cases:` generates — while an explicitly authored `:id:` always wins. Suite IDs are unaffected.

## Two things that bite on adoption, both documented

- **The IDs are lowercase**, which the sphinx-needs default `needs_id_regex` (`^[A-Z0-9_]{5,}`) rejects for local needs. It has to be relaxed, e.g. to `^[A-Za-z0-9_-]{6,}`. External needs skip the check entirely, which is why a generator producing external needs never hits this.
- **googletest parameterised suites put `/` into class and case names** (`ParamTest/0`), which is neither a legal need ID nor a usable HTML anchor. The readable part of the ID is reduced to `[A-Za-z0-9_]`, while the digest is taken over the raw name — so IDs stay byte-identical to the scheme's wherever the name is already ID-safe, and are valid where it would otherwise produce an unusable one.

## Verification

22 new tests: unit tests pinning the hash, display name and ID format against literal expected values, plus a Sphinx fixture project asserting the five expected IDs render and that the previous parent-scoped case fragment is gone. 104 passing, mypy strict and pre-commit clean, docs build clean with `-W`.


docs/configuration.rst
```diff
@@ -161,6 +161,37 @@ See :ref:`tr_source_file_option`.
...
+the ID of a test case changes when the test starts failing differently. That
+breaks anything holding a reference to it: build caches, a link authored from a
+test specification to its expected result, and diffs of an exported needs.json.
+
+With this option enabled, IDs have the form
+``{type}__{Classname__Casename}_{hash}``, where the hash is derived from the
+source file and the case name. Both ID-producing paths honour it: the
+``test-case`` directive and the IDs generated by ``:auto_cases:``. An explicitly
+authored ``:id:`` always wins. Test-suite IDs are unaffected.
+
```
MaximilianSoerenPollak, line 183: How does a realistic hash then look like? Just as an example for me.

MaximilianSoerenPollak (approved):
Code looks good as far as I can judge. 

I just had one question for my understanding.

docs/configuration.rst
ubmarco: Good question — here's a realistic one. For a pytest case in `tests/test_math.py` with class `TestMath` and name `test_addition`,the ID becomes:

```
testcase__TestMath__test_addition_oansv
```

The hash (`oansv`) is the first five lowercase base32 letters of the SHA-256 digest of `tests/test_math.pyTestMath__test_addition` (the source file concatenated with the display name `Classname__Casename`). I've also added this recipe and an example to the docs.

PR #145, opened 2026-09-04
feat(config): declarative [test_reports] section in ubproject.toml

ubmarco:
Stacked on #150 (sphinx-needs 6.0.1 / Sphinx 7.4 floor) and, through it, on the Python 3.11 floor of #147 — reading the section needs `tomllib`.

Describe the project once, in the `[test_reports]` section of `ubproject.toml` — the declarative file sphinx-needs, sphinx-codelinks, sphinx-mounts and ubCode already read — instead of restating it in `conf.py`.

**Rescoped.** This PR previously also carried the `test-reports convert` CLI integration and the conversion settings themselves. #144 has been reverted (#146) so the declarative configuration lands first; the converter returns as a follow-up PR stacked on this one, bringing its own settings under `[test_reports.build.needs]` and reading them from the start rather than shipping flags and growing `--config` afterwards. What is left here is exactly what the Sphinx extension consumes.

## What

- New Sphinx-free module `projectconfig.py` parses, validates and normalises the section. It imports no Sphinx, and neither does the package `__init__` it runs through: `setup` is resolved lazily (PEP 562), so a build action can read the section without the documentation toolchain installed. The section describes the *project*, not this extension.
- **Bridge**: a `config-inited` handler at `priority=100` applies the Sphinx-facing keys to their `tr_*` config values, ahead of the handlers that read them. Precedence is **`-D` > `ubproject.toml` > `conf.py` > built-in default** — the file is the source of truth for the project, the command line stays the per-invocation escape hatch. Both are announced in the log (`Applied … from …`, `Kept the -D value of …`).
- **New confval `tr_config_from_toml`** (default `ubproject.toml`; `None` disables, and `NoneType` is an accepted type so the documented opt-out does not trip Sphinx's own confval check). With the default name the file is searched for upwards from the `confdir` to the repository root — the directory holding `.git`, and nothing else bounds the search: a `pyproject.toml` on the way up marks a distribution, not the project, so `docs/pyproject.toml` beside `conf.py` and a uv-workspace member (`packages/<dist>/docs/conf.py` with one `ubproject.toml` at the monorepo root) both find the file. A fruitless search says where it ended (`sphinx-build -v`). An explicitly named file is used as-is and warns if missing (`test_reports.missing_config`).
- **Spellings**: `file`/`suite`/`case` accept both the positional `conf.py` list and a named table (`directive`/`type`/`name`/`prefix`/`color`/`style`). Relative `rootdir`/`report_template` anchor at the TOML file's directory; the loader joins without resolving, so the form of the path it is handed (Sphinx's already-resolved `confdir`, a converter's working directory) is kept as is.
- **Error policy**: a *known* key with the wrong type is an error — that is the typo class this validation exists for — and table *values* are checked, not just the outer table, including the values inside a `file`/`suite`/`case` table. An *unknown* key is warned about and ignored (suppressible via `test_reports.unknown_key`): the file is shared with tools on independent release cadences, so a key this version does not model must not take a build down. Same posture sphinx-mounts documents for `[[source.mounts]]`.
- **Foreign sub-table**: `[test_reports.build]` — and only that one — is recognised and passed through untouched. It holds the settings of the `test-reports build` command line, one sub-table per artifact it produces (`[test_reports.build.needs]` for turning test reports into a `needs.json` outside Sphinx), none of which this extension does, so it must draw neither an unknown-key warning nor a `tr_*` value. Any other sub-table is an unknown key like any other. The converter PR defines and validates the contents of `build.needs`.
- `tr_rootdir` declares `types=(str, os.PathLike)`: its default is Sphinx's `confdir`, a `_StrPath`, so a plain string — the only thing TOML or a string literal in `conf.py` can supply — used to fail Sphinx's confval type check and take a `-W` build down. Pre-existing, but the documented example newly advertised it.

## Notes

- The section is spelled `test_reports`, matching every other section of `ubproject.schema.json` (`build_tags`, `format_rst`, `needs_json`, …) and leaving the existing `[reports]` — report *templates* — unambiguous.

## Tests

54 tests in `tests/test_project_config.py`: loader validation, normalisation and path anchoring; upward discovery, its repository-root boundary and the layouts a `pyproject.toml` must not end (`docs/pyproject.toml`, workspace member); the foreign-sub-table passthrough; real Sphinx builds asserting the precedence chain (`-D` beats TOML, TOML beats `conf.py`), the warning paths, where a fruitless search ended, and that the documented example — read from `configuration.rst` verbatim — applies without a warning; and that `projectconfig` imports in a process where Sphinx is blocked.

## Review

Reviewed in https://github.com/useblocks/sphinx-test-reports/pull/145#issuecomment-5551256944, addressed in https://github.com/useblocks/sphinx-test-reports/pull/145#issuecomment-5551355689. The CLI-side fixes from that round move with the converter to its own PR.

Second round: https://github.com/useblocks/sphinx-test-reports/pull/145#pullrequestreview-5126626676 — the three fixes, the text corrections and both unconfirmed claims are addressed in 809bd9a; the discovery-rule question across packages stays open for the monorepo.

Two findings from the first review move with the conversion settings and are re-applied in the converter PR: the `need_type` / `case.type` agreement check (both name the need type of a test case, so they must not disagree), and the `link_properties` value-type validation.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1,5 +1,7 @@
 :hide-navigation:
 
+.. _configuration:
+
 Configuration
 =============
 The following options can be set inside the ``conf.py`` file of your Sphinx project.
@@ -401,3 +403,88 @@ An example of a JSON file, which supports the below configuration, can be seen i
          }
       }
    }
+
+Declarative configuration (ubproject.toml)
+------------------------------------------
+.. versionadded:: 1.5.0
+
+All of the above can also be configured declaratively, in the
+``[test_reports]`` section of your project's ``ubproject.toml`` -- the same
+shared file other useblocks tooling (sphinx-needs, sphinx-codelinks,
+sphinx-mounts, ubCode) reads. It describes the project once, so every tool
+acting on it works from the same settings instead of each restating them.
+
+.. code-block:: toml
+
+   [test_reports]
+   file_option = "report_file"
+   source_file_option = "file"
+   source_line_option = "line"
+   deterministic_case_ids = true
+   extra_options = ["more_info", "priority"]
+   property_link_types = { request = "req" }
+   rootdir = "docs"
+
+   # Need types: named tables (recommended) ...
+   [test_reports.case]
+   directive = "test-case"
+   type = "testcase"
+   name = "Test-Case"
+   prefix = "TC_"
+   color = "#999999"
+   style = "rectangle"
+
+   # ... or the positional list spelling of conf.py:
+   # case = ["test-case", "testcase", "Test-Case", "TC_", "#999999", "rectangle"]
+
+**Keys.** Every key is named like its ``tr_*`` config value without the prefix
+(``file_option`` configures ``tr_file_option``, and so on). A key carrying the
+wrong type is an error -- that is the typo class this validation exists to
+catch. An *unknown* key is reported as a warning and ignored: the file is
+shared with tools on independent release cadences, so a key this version does
+not model must not take your build down. Add ``"test_reports.unknown_key"`` to
+Sphinx's ``suppress_warnings`` to silence that report in a project that builds
+with ``-W``.
+
+Sub-tables belonging to other tools are left alone. Settings for converting
+test reports into a ``needs.json`` outside Sphinx live under
+``[test_reports.convert]``, which this extension does not read.
+
+**Precedence.** ``-D`` on the ``sphinx-build`` command line beats the TOML
+file, which beats ``conf.py``, which beats the built-in default. The
+declarative file is the source of truth for the project; the command line stays
+the per-invocation escape hatch.
+
+**Paths.** Relative values of ``rootdir`` and ``report_template`` are resolved
+against the directory containing the TOML file (not against ``conf.py`` or the
+working directory), so both consumers resolve them identically and the file
+stays self-describing when moved as a unit.
+
+**Deterministic IDs.** A build that imports a ``needs.json`` carrying
+deterministic case IDs, next to locally created test-case needs, must set
+``deterministic_case_ids = true`` so both ID schemes agree.
+
+.. _tr_config_from_toml:
+
+tr_config_from_toml
+~~~~~~~~~~~~~~~~~~~
+.. versionadded:: 1.5.0
+
+Name of the declarative configuration file whose ``[test_reports]`` section is
+applied to the ``tr_*`` values above. Defaults to ``ubproject.toml``.
+
+With the default name, the file is searched for in your ``confdir`` and its
+parent directories, stopping at the project root (a directory holding ``.git``
+or ``pyproject.toml``). That is what lets the canonical layout work -- the
+shared ``ubproject.toml`` at the repository root, ``conf.py`` in ``docs/`` --
+and lets a tool started anywhere below the root find exactly the same file by
+searching upward in the same way. A missing default file is not an error.
+
+Set to any other value to name a file explicitly; it is resolved against the
+``confdir``, is not searched for, and a warning is emitted if it does not
+exist. Set to ``None`` to switch declarative configuration off entirely.
+
+.. code-block:: python
+
+   tr_config_from_toml = "../ubproject.toml"   # explicit path
+   tr_config_from_toml = None                  # disable
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -442,13 +442,19 @@ acting on it works from the same settings instead of each restating them.
 wrong type is an error -- that is the typo class this validation exists to
 catch. An *unknown* key is reported as a warning and ignored: the file is
 shared with tools on independent release cadences, so a key this version does
-not model must not take your build down. Add ``"test_reports.unknown_key"`` to
-Sphinx's ``suppress_warnings`` to silence that report in a project that builds
-with ``-W``.
+not model must not take your build down.
 
-Sub-tables belonging to other tools are left alone. Settings for converting
-test reports into a ``needs.json`` outside Sphinx live under
-``[test_reports.convert]``, which this extension does not read.
+One sub-table belongs to another tool and is left alone: the settings for
+converting test reports into a ``needs.json`` outside Sphinx live under
+``[test_reports.convert]``, which this extension does not read. Any other
+sub-table is treated like any other unknown key -- reported and ignored.
+
+**Warnings.** The two warnings this feature emits carry a type, so either can
+be silenced through Sphinx's ``suppress_warnings`` in a project that builds
+with ``-W``: ``test_reports.unknown_key`` for the unknown-key report above, and
+``test_reports.missing_config`` for an explicitly named file that does not
+exist (see :ref:`tr_config_from_toml`). A known key with the wrong type is an
+error, not a warning, and cannot be suppressed.
 
 **Precedence.** ``-D`` on the ``sphinx-build`` command line beats the TOML
 file, which beats ``conf.py``, which beats the built-in default. The
@@ -474,15 +480,22 @@ Name of the declarative configuration file whose ``[test_reports]`` section is
 applied to the ``tr_*`` values above. Defaults to ``ubproject.toml``.
 
 With the default name, the file is searched for in your ``confdir`` and its
-parent directories, stopping at the project root (a directory holding ``.git``
-or ``pyproject.toml``). That is what lets the canonical layout work -- the
-shared ``ubproject.toml`` at the repository root, ``conf.py`` in ``docs/`` --
-and lets a tool started anywhere below the root find exactly the same file by
-searching upward in the same way. A missing default file is not an error.
+parent directories, up to the repository root (the directory holding ``.git``).
+That is what lets the canonical layout work -- the shared ``ubproject.toml`` at
+the repository root, ``conf.py`` in ``docs/`` -- and lets a tool started
+anywhere below the root find exactly the same file by searching upward in the
+same way. Only the repository root bounds the search: a ``pyproject.toml`` on
+the way up does not, so a ``docs/`` directory with its own ``pyproject.toml``
+and the documentation of a workspace member in a monorepo
+(``packages/<name>/docs/conf.py``) both find the file at the root. A missing
+default file is not an error; ``sphinx-build -v`` reports where the search
+ended.
 
 Set to any other value to name a file explicitly; it is resolved against the
-``confdir``, is not searched for, and a warning is emitted if it does not
-exist. Set to ``None`` to switch declarative configuration off entirely.
+``confdir``, is not searched for, and a warning of type
+``test_reports.missing_config`` is emitted if it does not exist -- the build
+then runs on the ``conf.py`` configuration. Set to ``None`` to switch
+declarative configuration off entirely.
 
 .. code-block:: python
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -9,6 +9,8 @@ The following options can be set inside the ``conf.py`` file of your Sphinx proj
 .. contents::
    :local:
 
+.. _tr_rootdir:
+
 tr_rootdir
 ----------
 ``tr_rootdir`` takes a path, which is used as *root dir* for all provided file paths in other directives.
@@ -72,6 +74,8 @@ By default ``tr_case`` is set to::
    ['test-case', 'testcase', 'test-case', 'TC_', '#999999', 'node']
 
 Please read :ref:`tr_file` for more details.
+
+.. _tr_report_template:
 
 tr_report_template
 ------------------
@@ -423,6 +427,8 @@ acting on it works from the same settings instead of each restating them.
    deterministic_case_ids = true
    extra_options = ["more_info", "priority"]
    property_link_types = { request = "req" }
+   # Base directory the directives look their report files up under -- not an
+   # output directory. Relative to this file, see Paths below.
    rootdir = "docs"
 
    # Need types: named tables (recommended) ...
@@ -462,10 +468,15 @@ file, which beats ``conf.py``, which beats the built-in default. The
 declarative file is the source of truth for the project; the command line stays
 the per-invocation escape hatch.
 
-**Paths.** Relative values of ``rootdir`` and ``report_template`` are resolved
-against the directory containing the TOML file (not against ``conf.py`` or the
-working directory), so both consumers resolve them identically and the file
-stays self-describing when moved as a unit.
+**Paths.** ``rootdir`` (:ref:`tr_rootdir`) is the directory the relative report
+paths in the directives are resolved against -- with ``rootdir = "docs"``,
+``.. test-file:: reports/pytest.xml`` reads ``docs/reports/pytest.xml``. It is
+an input location, not an output directory: nothing is written there.
+``report_template`` (:ref:`tr_report_template`) names a custom template file.
+Relative values of both are resolved against the directory containing the TOML
+file (not against ``conf.py`` or the working directory), so both consumers
+resolve them identically and the file stays self-describing when moved as a
+unit.
 
 **Deterministic IDs.** A build that imports a ``needs.json`` carrying
 deterministic case IDs, next to locally created test-case needs, must set
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -444,10 +444,11 @@ catch. An *unknown* key is reported as a warning and ignored: the file is
 shared with tools on independent release cadences, so a key this version does
 not model must not take your build down.
 
-One sub-table belongs to another tool and is left alone: the settings for
-converting test reports into a ``needs.json`` outside Sphinx live under
-``[test_reports.convert]``, which this extension does not read. Any other
-sub-table is treated like any other unknown key -- reported and ignored.
+One sub-table belongs to another tool and is left alone: ``[test_reports.build]``
+holds the settings of the ``test-reports build`` command line, one sub-table per
+artifact it produces -- ``[test_reports.build.needs]`` for turning test reports
+into a ``needs.json`` outside Sphinx -- none of which this extension does. Any
+other sub-table is treated like any other unknown key -- reported and ignored.
 
 **Warnings.** The two warnings this feature emits carry a type, so either can
 be silenced through Sphinx's ``suppress_warnings`` in a project that builds
--- a/sphinxcontrib/test_reports/projectconfig.py
+++ b/sphinxcontrib/test_reports/projectconfig.py
@@ -1,0 +1,397 @@
+"""Declarative project configuration for sphinx-test-reports.
+
+Reads the ``[test_reports]`` section of a project's ``ubproject.toml`` -- the
+declarative file shared with the other useblocks tooling (sphinx-needs,
+sphinx-codelinks, sphinx-mounts, ubCode) -- so that a project is described once
+instead of being restated in every tool that acts on it.
+
+**Nothing in this module may import Sphinx.** The section describes the project,
+not this extension, and a report-to-``needs.json`` converter has to be able to
+read it as a build action without the documentation toolchain installed.
+
+The keys are the Sphinx-facing configuration (:data:`BRIDGE_KEYS`), spelled
+like the ``tr_*`` config values without the prefix. The section may also carry
+sub-tables for tools other than the Sphinx extension -- a report converter
+reads ``[test_reports.convert]`` -- which this reader does not interpret.
+
+**Error policy.** A known key carrying the wrong type is fatal: that is the
+typo class this validation exists to catch, and letting it through would
+silently change what gets produced. An *unknown* key is only reported. The
+file is shared with tools on independent release cadences, so a key this
+reader does not model is routine rather than a mistake -- and aborting on it
+would take down every build of the project on every older sphinx-test-reports,
+including builds the key would not have changed. This is the same posture
+sphinx-mounts takes for ``[[source.mounts]]``.
+"""
+
+import tomllib
+from pathlib import Path
+from typing import Callable, Mapping, NoReturn, Sequence
+
+#: Default file the configuration is read from. Looked up by walking up from
+#: the ``confdir`` (Sphinx) or the working directory (a converter); see
+#: :func:`find_project_config`. ``ubproject.toml`` is the convention shared
+#: with other useblocks tooling so a single declarative file describes the
+#: project to every downstream consumer -- Sphinx, ubCode, a converter --
+#: without
+#: any of them having to execute Python.
+DEFAULT_TOML_FILENAME = "ubproject.toml"
+
+#: The section this extension owns inside the shared file. Spelled
+#: ``snake_case`` like every other section of ``ubproject.schema.json``
+#: (``build_tags``, ``format_rst``, ``needs_json``, ``rst_lint``, ...); note
+#: that ``[reports]`` is already taken, and means report *templates*.
+SECTION = "test_reports"
+
+#: Directory entries that end the upward search of
+#: :func:`find_project_config`. A directory carrying one of these is a project
+#: root, so a consumer below it must not silently adopt the configuration of
+#: an unrelated parent project. ``ubproject.toml`` itself is not listed --
+#: finding it is the success case, checked first.
+_ROOT_MARKERS = (".git", "pyproject.toml")
+
+#: Section keys bridged onto their ``tr_*`` Sphinx config values.
+BRIDGE_KEYS = (
+    "file",
+    "suite",
+    "case",
+    "file_option",
+    "source_file_option",
+    "source_line_option",
+    "rootdir",
+    "report_template",
+    "suite_id_length",
+    "case_id_length",
+    "import_encoding",
+    "extra_options",
+    "property_link_types",
+    "json_mapping",
+    "deterministic_case_ids",
+)
+
+#: Keys holding a path. Relative values are anchored against the directory
+#: containing the TOML file, not against ``confdir`` or the process working
+#: directory: the file is self-describing, and moving it as a unit keeps its
+#: relative paths meaningful. It also makes every consumer resolve a
+#: relative path identically.
+PATH_KEYS = ("rootdir", "report_template")
+
+#: The keys whose need-type setting accepts both the positional ``conf.py``
+#: list and a named table; :func:`_normalise_type_entry` reduces both to the
+#: list form. ``None`` in :data:`_KEY_TYPES` marks exactly these.
+_DUAL_SPELLING_KEYS = ("file", "suite", "case")
+
+#: Sub-tables of the section that belong to another consumer. They are
+#: recognised so they do not draw an unknown-key warning, and deliberately not
+#: interpreted: ``convert`` holds the settings for turning test reports into a
+#: ``needs.json`` outside Sphinx, which this extension never does.
+FOREIGN_TABLES = ("convert",)
+
+#: Expected Python type per key. ``bool`` must be checked *before* ``int``
+#: (bool is an int subclass). ``None`` marks the dual-spelling keys, which are
+#: normalised separately; path keys are validated as ``str`` and anchored
+#: afterwards.
+_KEY_TYPES: dict[str, type[object] | None] = {
+    "file": None,
+    "suite": None,
+    "case": None,
+    "file_option": str,
+    "source_file_option": str,
+    "source_line_option": str,
+    "rootdir": str,
+    "report_template": str,
+    "suite_id_length": int,
+    "case_id_length": int,
+    "import_encoding": str,
+    "extra_options": list,
+    "property_link_types": dict,
+    "json_mapping": dict,
+    "deterministic_case_ids": bool,
+    "convert": dict,
+}
+
+#: Required type of the *values* inside a table-valued key. Without this a
+#: table passes validation on its outer shape alone, and a mistake such as
+#: ``property_link_types = { request = ["req"] }`` reaches the directives,
+#: which fail with a bare ``TypeError`` on an unhashable field name.
+#: ``None`` means the nested shape is free-form -- ``json_mapping`` mirrors an
+#: arbitrary parser mapping -- so only the outer table is checked.
+_DICT_VALUE_TYPES: dict[str, type[object] | None] = {
+    "property_link_types": str,
+    "json_mapping": None,
+    "convert": None,
+}
+
+#: Field order of the positional ``tr_file``-style lists, and the table keys
+#: that spell the same thing readably.
+_TYPE_ENTRY_FIELDS = ("directive", "type", "name", "prefix", "color", "style")
+
+#: ``tomllib.TOMLDecodeError`` bound through an annotation: the attribute
+#: expression itself is typed loosely enough to trip the strict ``Any`` bans,
+#: and an ``except`` clause has no annotation of its own to absorb it.
+_TOML_DECODE_ERROR: type[Exception] = tomllib.TOMLDecodeError
+
+
+class TomlConfigError(Exception):
+    """Raised when the declarative config cannot be parsed or is malformed.
+
+    Deliberately *not* a ``SphinxError``: this module must stay importable
+    without Sphinx. The Sphinx-side bridge re-raises it as an
+    ``InvalidConfigurationError`` to abort the build; a non-Sphinx consumer
+    reports it and exits non-zero.
+    """
+
+
+def find_project_config(
+    start: Path, filename: str = DEFAULT_TOML_FILENAME
+) -> Path | None:
+    """Search *start* and its parents for *filename*.
+
+    The declarative file conventionally sits at the project root while its
+    consumers run from below it -- ``conf.py`` in ``docs/``, a build action
+    from wherever CI invoked it -- so anchoring strictly at the caller's own
+    directory would leave the shared file unread by one of them, silently.
+
+    The walk stops at the first directory holding *filename*, and at a
+    directory that looks like a project root (:data:`_ROOT_MARKERS`) even when
+    it does not hold the file, so a consumer never adopts the configuration of
+    an unrelated parent project.
+
+    :return: The file, or ``None`` when the search reached a project root or
+        the filesystem root without finding one.
+    """
+    for directory in (start, *start.parents):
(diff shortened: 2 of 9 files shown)


ubmarco:
## Review

**CI is red on this head.** All Python 3.10 jobs fail and `mypy` reports 63 errors, so `all_good` is failing — the 3.12 jobs were only *cancelled* by fail-fast, they pass on their own. Both root causes are introduced here (1 and 2).

Measured against how `sphinx-codelinks` and `sphinx-mounts` read the same shared file, this reader is the **strictest of the three and the only one enabled by default**, which is where most of the design risk sits (7, 8, 9).

No prior human reviews on this PR, so nothing outstanding to re-check.

---

### 1. [High] `import tomllib` breaks the extension on Python 3.10

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/projectconfig.py#L33-L38

`tomllib` is stdlib only from 3.11, but `requires-python = ">=3.10"`, `noxfile.py` has `PYTHON_VERSIONS = ["3.10", "3.12"]`, and CI runs the matrix on 3.10. `test_reports.py` imports `projectconfig` at module level, so this is not confined to the CLI: *every* `sphinx-build` of a project with this extension dies with `ModuleNotFoundError: No module named 'tomllib'` on the advertised minimum interpreter. The ten red 3.10 jobs are all this.

Either add the `tomli` backport, or bump `requires-python` to `>=3.11` and drop 3.10 from nox/CI/classifiers — `sphinx-codelinks` and `sphinx-mounts` are both already `>=3.12`. The classifier list also still advertises 3.9.

### 2. [High] `projectconfig.py` fails the repo's strict mypy gate

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/projectconfig.py#L86-L92

`[tool.mypy]` is `strict = true` with `disallow_any_expr` and `disallow_any_explicit`. The new file is in scope and not excluded, so `mypy` exits 1 with 63 errors, nearly all `Expression has type "Any"` / `Explicit "Any" is not allowed` off `_KEY_TYPES: dict[str, Any]`, plus an `Unused "type: ignore"` at `projectconfig.py:193`.

Worth fixing rather than adding an exclude entry — that list is an explicit "TODO: reactivate" legacy allowance. The house style for the strict-checked modules is already `object` + `isinstance` narrowing (see `_optional`/`_mappings` in `needs_export.py`), and both sibling repos model the section as typed structures instead of an untyped dict: codelinks uses dataclasses carrying a JSON-Schema fragment in `field(metadata={"schema": …})`, sphinx-mounts frozen dataclasses with `from_dict`. That also replaces the hand-rolled `_KEY_TYPES`/`_type_label` dispatch and gives you a schema fragment to contribute to `ubproject.schema.json`.

### 3. [High] `-D` overrides are silently discarded, and the docstring claims the opposite

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/test_reports.py#L244-L250

> Consumers that need per-invocation overrides still have the `-D` command line, which Sphinx applies after this event.

Backwards. Sphinx materialises `-D` *before* `config-inited` in every supported version (`Config.init_values()` on 4.x–8.x; `Config.__getattr__` consulting `_overrides` and caching into `__dict__` on 9.x), so the `setattr` at `sphinxcontrib/test_reports/test_reports.py:269` overwrites them with no warning. Reproduced against the PR's own fixture:

```
conf.py value   : confpy_report_file
ubproject.toml  : report_file
-D override     : cli_override_wins
RESULT          : report_file      <-- the -D override is lost
```

The sharp edge is `-D tr_deterministic_case_ids=1` — the documented way to switch a CI build onto the ID scheme the convert CLI always writes. If `ubproject.toml` sets that key, the flag is ignored, the build mints content-hashed IDs against a needs.json carrying deterministic ones, and the cross-links dangle. Fix: skip keys present in the overrides (or at minimum warn), and correct the docstring. No test covers `-D`.

### 4. [High] Sphinx anchors at `confdir`, the CLI at `cwd` — the canonical layout gets no config

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/cli.py#L174-L181

Neither side searches upward. For the layout that motivates the feature — `repo/ubproject.toml` (the file sphinx-needs and ubCode already read) plus `repo/docs/conf.py` — `test-reports convert` at the repo root applies the section, but `sphinx-build docs _build` looks for `docs/ubproject.toml`, finds nothing, and applies none of it **silently**. A directive using an option declared through `extra_options` then fails with `unknown option: more_info`, and report paths land on `file` instead of the declared `report_file` — the two consumers describe different fields, which is the exact divergence this PR exists to remove.

`sphinx-codelinks` already does root discovery by walking up for `.git` / `ubproject.toml` / `pyproject.toml` (`analyse/preproc/compile_db.py:12`).

### 5. [High] Dict-valued keys are never value-type-checked

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/projectconfig.py#L186-L193

Item validation is gated on `expected is list`, so `link_properties`, `property_link_types` and `json_mapping` are checked for being a table and nothing more — the "wrong types are an error" promise does not hold for them. `link_properties = { PartiallyVerifies = ["partially_verifies"] }` (an easy mistake next to the array-valued `tags`) loads clean, and `cli.py:222` then `str()`-coerces it, so `convert` exits 0 having written a field literally named `"['partially_verifies']"` onto every need, with no `partially_verifies` link field — silent corruption of the artifact a downstream `needimport` consumes. The same gap sends `property_link_types = { request = ["req"] }` into `directives/test_common.py:212`, where the build dies with a raw `TypeError: unhashable type: 'list'` from inside a directive instead of an `InvalidConfigurationError`.

### 6. [High] The documented `tr_config_from_toml = None` opt-out trips Sphinx's own confval check

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/test_reports.py#L124-L128

`add_config_value` gets no `types`, so Sphinx infers `str` from the default and `check_confval_types` warns on the documented opt-out. Reproduced:

```
WARNING: The config value `tr_config_from_toml' has type `NoneType', defaults to `str'.
```

Any project building with `-W` fails outright by following the documented way to disable the feature. Fix is `types=(str, type(None))`.

### 7. [Medium] `need_type` and `[test-reports.case].type` both set the need type, with no cross-check

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/projectconfig.py#L48-L57

The CLI takes its type and deterministic-ID prefix from `need_type`; the build takes them from `tr_case[1]` (`directives/test_common.py:98`). Both default to `testcase` but come from independent keys in the same file. Rename the case type only in the recommended named table and `convert` still emits `testcase__…` IDs while the build registers `check` and mints `check__…`: `needimport` reports an unknown need type and the imported and locally generated cases for the same test carry different IDs. For a file whose purpose is keeping the two consumers in agreement, this pair should be validated as agreeing (or derived from one another).

### 8. [Medium] Default-on plus fatal unknown keys is a forward-compat hazard on a *shared* file

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/projectconfig.py#L170-L176

The PR's own rationale is that this file is read by several tools on independent release cadences — which argues against aborting on a key this version does not model. `tr_config_from_toml` defaulting to `ubproject.toml` maximises the blast radius: today a newer or foreign `[test-reports]` table is inert, after this PR it is fatal on every older install, including builds the key would not have changed.

Both siblings go the other way, and `sphinx-mounts` documents exactly this reasoning in `MountConfig.from_dict`:

> Reported, never fatal. A `ubproject.toml` is shared with tools on independent release cadences, so a key this reader does not model is routine rather than a mistake — and aborting on it takes down every build of the project on every older sphinx-mounts […]

`sphinx-codelinks` additionally defaults `config_from_toml = None` and only warns on a parse failure. Keeping hard errors for a *known* key with a wrong type preserves the typo protection the message argues for; unknown keys do not need to be fatal.

### 9. [Medium] `[test-reports]` breaks the ubproject.toml section-naming convention

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/projectconfig.py#L41-L46

Every section in the published `ubproject.schema.json` is one lowercase word or snake_case: `build_tags`, `codelinks`, `extend`, `format_rst`, `lint`, `needs`, `needs_json`, `parse`, `project`, `quality`, `reports`, `rst_lint`, `scripts`, `server`, `source`, `virtual`, `workflow`. `test-reports` would be the only hyphenated one and needs an explicit serde `rename` whenever ubCode models it (`UbprojectConfig`, `rust/ubc_config/src/top.rs`). **`[reports]` also already exists there**, meaning report-*template* configuration — easy to confuse. Renaming after release is breaking, so it is cheaper to settle now.

(No breakage today: `UbprojectConfig` has no `deny_unknown_fields` and the schema sets no top-level `additionalProperties`, so an unmodelled table is tolerated — it just gets no validation or completion.)

### 10. [Medium] A mistyped `tr_config_from_toml` is swallowed without a diagnostic

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/test_reports.py#L255-L263

`load_project_config` returns `None` for a missing file and `if not section: return` treats that as "no section". `tr_config_from_toml = "ubprojet.toml"` therefore builds completely normally on the conf.py values — field renames and extra options vanish, every need carries the wrong field names, nothing points at the cause. `docs/cli.rst` promises "an explicitly given path must exist" and `cli.py:175` enforces it for `--config`; `sphinx-codelinks` warns in this case.

### 11. [Medium] `OSError` from the config read escapes as a raw traceback

https://github.com/useblocks/sphinx-test-reports/blob/58208598665d4e40bf505241ddc40e7e55a911ae/sphinxcontrib/test_reports/projectconfig.py#L143-L151

Only `TOMLDecodeError` is wrapped, but `path.open("rb")` can still raise after `is_file()` succeeded — unreadable permissions, or the file replaced between the check and the open. `_load_config` catches only `TomlConfigError`, so `convert` aborts with a `PermissionError` traceback rather than the `error: …` / exit 2 its docstring promises, and the build ends in an unhandled traceback instead of an `InvalidConfigurationError`.

### Low

- **`--remote-url`/`--commit` guard names flags that were never passed.** Moved from `arguments` to the merged `settings` (`cli.py:244`), but the message still names only the flags — setting just `remote_url` in the TOML yields `error: --remote-url and --commit must be given together` for an invocation with no flags at all. `test_toml_half_remote_pair_is_rejected` asserts the exit code but not the message.
- **`CONVERSION_KEYS` is dead** — exported and documented but never read; `_resolve_config` re-enumerates the same eight keys by hand, so the two can drift. Similarly `BRIDGE_KEYS`/`_KEY_TYPES` hand-mirror the `add_config_value` registry with no test asserting they stay in sync, the dual-spelling key set is spelled twice (`None` entries in `_KEY_TYPES` and again as a literal tuple), link-property normalisation is implemented in both `_resolve_config` and `_parse_link_properties`, and the config file's existence is probed three times across `_load_config` and `load_project_config`.
- **No changelog entry.** Every sibling PR in this series (#141–#144) added a bullet to `Unreleased` in `docs/changelog.rst`; this one does not. The `.. versionadded:: 1.5.0` markers are consistent with the rest of the series.
- **`tr_config_from_toml` has no confval section** in `configuration.rst` — it appears only in prose, though it is the gate for the whole feature and defaults to on.
- **The stated reason for avoiding `app.connect(priority=…)` is wrong.** The PR body says it "needs Sphinx ≥ 7.4"; `Sphinx.connect(event, callback, priority=500)` has accepted `priority` since **Sphinx 3.0**, and both siblings use it (codelinks `priority=10`, sphinx-mounts `priority=450`). An explicit priority makes the ordering contract the comment describes enforceable instead of a convention a future reorder breaks silently.
- **`test_broken_explicit_config_is_an_error` never passes `--config`**, so the explicit-path error branch it names is untested.
- **`_anchor_paths` calls `.resolve()`**, collapsing symlinks that the `conf.py` spelling of the same option preserves.

---

*Correctness pass: 4 finder angles → 39 candidates → 27 independent verifiers; 2 refuted and excluded.*

**Agreed follow-up** (fixes coming as a separate commit on this branch): drop Python 3.10 (already out of support) rather than adding the `tomli` backport; unknown keys become a warning, never fatal, with wrong types on known keys staying hard errors; the section is renamed to `[test_reports]` to match the ubproject.toml convention; and both consumers gain walk-up discovery so the repo-root file is found from `docs/`.

ubmarco:
All findings above are addressed in a658cfac8e72df88e31b81dd4b7058723df674fc. **CI is green: 26/26 checks pass**, against 4/26 on the reviewed head — the full 3.11/3.12 × Sphinx × sphinx-needs matrix (20 jobs), `mypy`, `pre-commit`, `Docs-Linkcheck`, coverage, `all_good` and Read the Docs.

The three design questions the review left open were settled as: **warn, never fatal** on unknown keys; **rename to `[test_reports]`**; **walk-up discovery**. The PR description has been updated to match.

## High

**1. `import tomllib` on Python 3.10** — dropped 3.10 rather than adding the `tomli` backport; it is out of upstream support. `requires-python = ">=3.11"` (`pyproject.toml:25`), `PYTHON_VERSIONS = ["3.11", "3.12"]` in `noxfile.py`, the CI matrix likewise, and the two stale 3.9 job drivers moved to 3.12. The 3.9/3.10 classifiers are gone.

**2. 63 mypy errors** — `projectconfig.py` retyped with `object` / `type[object]` instead of `Any`, which is the house style of the other strict-checked modules (`_optional`, `_mappings` in `needs_export.py`). The only concession is one annotated binding for `tomllib.load`, whose return is `dict[str, Any]` by typeshed; everything downstream of `_read_toml` is `object`. `mypy` now reports *Success: no issues found in 8 source files*, and the file stays out of the exclude list.

**3. `-D` silently clobbered** — precedence is now `-D` > TOML > conf.py > default. `_command_line_overrides` (`test_reports.py:251`) reads the override map — spelled `_overrides` on newer Sphinx, `overrides` before that — and the bridge skips those keys (`test_reports.py:323`), logging which TOML values were superseded rather than dropping them silently. The docstring that claimed the opposite is rewritten. Covered by `test_command_line_override_beats_toml`.

**4. confdir vs cwd** — new `find_project_config` (`projectconfig.py:177`) searches upward from the `confdir` (Sphinx) or the working directory (CLI), stopping at the first directory holding the file, and at a `.git`/`pyproject.toml` project root when it does not — so a build never adopts an unrelated parent project's config. The canonical root-`ubproject.toml` + `docs/conf.py` layout now works from both sides. Six tests in `TestDiscovery` plus `test_walks_up_from_the_confdir` and the CLI's `test_walks_up_to_the_project_root`.

**5. Unvalidated table values** — `_DICT_VALUE_TYPES` (`projectconfig.py:146`) declares the required value type per table key, enforced by `_check_table_values` (`projectconfig.py:312`). `link_properties` and `property_link_types` require strings; `json_mapping` stays free-form deliberately, since validating deeper would reject valid parser mappings — that is now stated in the code and covered by `test_json_mapping_nesting_stays_free_form`.

**6. `tr_config_from_toml = None` tripping the confval check** — `types=(str, type(None))` (`test_reports.py:136`). `test_disabling_toml_reading_is_not_a_type_warning` asserts the build produces no warning naming the confval, so a `-W` project can follow the documented opt-out.

## Medium

**7. `need_type` vs `case.type`** — `_check_need_type_agreement` (`projectconfig.py:330`) rejects disagreeing values with a message explaining that one drives the CLI and the other the build.

**8. Unknown keys fatal by default** — now reported through a caller-supplied `warn` callback and dropped (`projectconfig.py:272`), so the loader stays Sphinx-free while the extension emits a real Sphinx warning under `test_reports.unknown_key` and the CLI prints to stderr. Suppressible via `suppress_warnings`, which keeps it non-fatal even under `-W`. Wrong types on *known* keys are still hard errors, so the typo protection the original message argued for is intact.

**9. Section naming** — renamed to `[test_reports]` across code, fixture, tests and docs.

**10. Mistyped `tr_config_from_toml`** — an explicitly named file that does not exist now warns (`test_reports.py:298`) instead of silently building on the conf.py values. A missing *default* file stays silent, as before.

**11. `OSError` escaping** — `_read_toml` (`projectconfig.py:236`) wraps both `TOMLDecodeError` and `OSError` into `TomlConfigError`, so an unreadable or racing file produces the documented `error: …` / exit 2 and an `InvalidConfigurationError` respectively.

## Low

- `_pair_requirement` (`cli.py:270`) names the TOML keys and the file when that is where `remote_url`/`commit` came from, and the flags otherwise; asserted on the message, not just the exit code.
- `_resolve_config` (`cli.py:252`) is driven by `CONVERSION_KEYS`, so the constant is load-bearing and a new key cannot be added without a default. Link-property normalisation now lives only in `_parse_link_properties`, and `_load_config` probes the file once.
- Changelog entries added for the feature and the Python 3.10 drop.
- `tr_config_from_toml` has its own confval section in `configuration.rst`, documenting the search, the explicit-path behaviour and the `None` opt-out.
- `app.connect(..., priority=100)` spells the ordering contract explicitly, so reordering the `connect` calls cannot silently break it; the PR description no longer claims `priority` needs Sphinx 7.4.
- `test_broken_explicit_config_is_an_error` passes `--config` and asserts the message.
- `_anchor_paths` (`projectconfig.py:443`) joins without `resolve()`, so symlinks survive as they do in the conf.py spelling; `test_symlinked_directories_are_preserved` covers it.

Also added: `--no-config`, so a conversion can be made to depend only on its arguments.

The test file grew from 31 to **56 tests**. The two refuted candidates (`_wrong_type`'s return annotation, the defensive `getattr`) were not acted on, though the `getattr` is gone anyway now that the confval is read directly.

chrisjsewell (commented):
Reviewed at `66900479` by construction rather than by reading: every claim in the description was built and measured on two cells (Sphinx 9.1.0 and 8.1.3, sphinx-needs 8.5.0), the eleven findings of the earlier round were re-tested and all hold closed, and the gates are green here (150/150 tests, ruff, strict mypy with `projectconfig.py` in scope, zero-warning docs and `-W` fixture builds). The feature works, and in two places better than the description says: the `-D` override is announced in the log rather than silent, and the TOML route's diagnostics are clearer than the `conf.py` route's.

Nine of eleven claims confirmed. The precedence chain `-D` > `ubproject.toml` > `conf.py` > default holds on a real command line; `priority=100` really is ahead of every `tr_*` reader; the error-versus-warning policy and `suppress_warnings = ["test_reports.unknown_key"]` behave exactly as documented; the foreign `convert` table passes through untouched.

**Three things I would fix before merging, all small:**

1. **Upward discovery stops before it starts when `confdir` is itself a project root.** `find_project_config` tests the `.git`/`pyproject.toml` markers on the starting directory, so `docs/pyproject.toml` beside `conf.py` — and every uv-workspace member layout, `packages/<dist>/pyproject.toml` with `docs/conf.py` and a root `ubproject.toml` — returns `None`, silently. Measured on exactly that layout. That is the monorepo this package is about to move into, and the failure is the one the feature exists to remove: the Sphinx build and the converter describing different fields. Either apply the marker test from the second step of the walk onward, or at least log which directory's marker ended the search. Worth a `TestDiscovery` case for the `docs/pyproject.toml` shape.
2. **One mutation survives the whole suite.** Dropping the non-string check on values inside `[test_reports.case]` leaves all 150 tests green. One more parametrise case (`color = 999999`) closes it; the loader already produces the right message.
3. **The documentation example fails when followed.** The new block in `configuration.rst` carries `rootdir = "docs"`, and a `-W` project that copies it gets `The config value 'tr_rootdir' has type 'str', defaults to '_StrPath'`. The root cause pre-dates this PR, but the example newly advertises it: drop `rootdir` from the example, or give `tr_rootdir` `types=(str, os.PathLike)`.

**Two sentences in the description are not quite true of the code** (no code change needed, just the text): "sub-tables belonging to other tools are left alone" is true only of `convert` — `[test_reports.other]` draws an unknown-key warning, which is arguably right, but the sentence should say so; and it is 40 tests in `test_project_config.py`, not 50. Two more small ones: the comment at `test_reports.py:128` says the file "must exist" where the code warns and continues, and the `test_reports.missing_config` subtype is not in the docs' warning list.

**Two claims I could not confirm as stated:**

- "Relative `rootdir`/`report_template` anchor at the TOML file's directory, preserving symlinks the way the `conf.py` spelling does." In a real build Sphinx resolves `confdir` itself, so the two routes disagree byte for byte under a symlinked docs directory. The loader-level tests cannot see this. It is a low-stakes edge, but the sentence promises something a build does not deliver.
- "`projectconfig.py` imports no Sphinx." True of the module, but `import sphinxcontrib.test_reports.projectconfig` executes the package `__init__`, which imports Sphinx eagerly, so in an environment without Sphinx the import dies with `ModuleNotFoundError: No module named 'sphinx'`. The converter PR's premise needs a lazy `__init__` (or the module moved out from under it) before it is usable; not a blocker here, but worth knowing before #148.

**For after this PR, not in it:** three packages now read the same file with three discovery rules — sphinx-needs' `needs_from_toml` is confdir-relative with no search (a user has to write `../ubproject.toml`), this one searches upward, and sphinx-mounts has its own. One shared rule is a monorepo question, and I will take it there. And nothing in ubCode models `[test_reports]` yet, which wants a one-line issue.

Constructions and measurements for each point are in a report I can share; the mutations were all reverted and nothing in this branch was touched.

ubmarco:
Addressed in 809bd9a, on top of the restack onto #150 (#151 and #148 are rebased onto it). b60308b on top renames the foreign sub-table to `[test_reports.build]` on the author's request, so the reviewed `convert` passthrough is now `build`. Re https://github.com/useblocks/sphinx-test-reports/pull/145#pullrequestreview-5126626676.

**The three fixes**

1. **Discovery.** `pyproject.toml` is no longer a root marker; the search is bounded by the repository root (`.git`) and by nothing else. Applying the marker test from the second step onward would have fixed `docs/pyproject.toml` beside `conf.py` but not the workspace member, where `packages/<dist>/pyproject.toml` *is* the second step, so the marker had to go rather than move. A fruitless search now reports where it ended (`no ubproject.toml in <confdir> or its parents up to the repository root <dir> (holding .git)`), logged at verbose level so a project without the file stays quiet; the converter gets the same line behind `-v` (#148). A relative start directory is made absolute first. `TestDiscovery` covers both layouts, a nested-repository boundary and the report; `TestBridgePrecedence` covers `docs/pyproject.toml` and the `-v` line in a real build.
2. **Mutation.** Three cases in `test_need_type_values_must_be_strings`: `color = 999999` in the table, a non-string element in the positional list, and a short list. Each was checked against the mutation it closes: dropping the table check or the list filter fails exactly one case.
3. **Docs example.** `tr_rootdir` declares `types=(str, os.PathLike)`. The same warning hit a plain string in `conf.py` before this PR, so the fix covers both routes. `test_the_documented_example_applies_without_warnings` reads the block from `configuration.rst` verbatim and builds it, so the example cannot drift from the code again.

**Text.** The docs now say that only `[test_reports.build]` (settings of the `test-reports build` command, `build.needs` for the `needs.json`) is passed through and that any other sub-table is an ordinary unknown key; both warning types (`test_reports.unknown_key`, `test_reports.missing_config`) are listed; the comment in `test_reports.py` says an explicit file warns rather than "must exist"; the description is corrected (54 tests, the discovery rule, the two claims below).

**The two unconfirmed claims**

- Symlinks: confirmed, Sphinx resolves `confdir` before the bridge runs, so the claim was not true of a build. It is gone from the description, the `_anchor_paths` docstring and the test, which now only asserts that the loader does not itself resolve the directory it is handed.
- `projectconfig` importable without Sphinx: the package `__init__` resolves `setup` lazily (PEP 562), taken from #151 so the stack carries one version. An `ImportError` from the lazy import is re-raised as `ExtensionError` when Sphinx is importable, because Sphinx fetches `setup` with `getattr()` and would otherwise show a raw traceback for a missing sphinx-needs. `test_projectconfig_imports_without_sphinx` imports the module in a subprocess with `sphinx` blocked.

Left as you framed them: the cross-package discovery rule and the ubCode model of `[test_reports]`, both for the monorepo.

chrisjsewell (commented):
Round 2, same posture as round 1: every fix reproduced by construction on three cells (sphinx 9.1 / needs 8.5.0, sphinx 8.1.3 / needs 6.3.0, and the new floor sphinx 7.4.7 / needs 6.0.1). 164/164, ruff, mypy, a zero-warning docs build and a `-W` build of the fixture all green here as in CI. The rebase is byte-for-byte: `8bde9fda..66900479` and `29c0f85..c569812` differ in six hunk offsets and nothing else.

**Closed, and verified as closed:** F1 (`tr_rootdir` accepts a path on both routes; the docs example is built verbatim by a test), F2 (four real builds: `docs/pyproject.toml`, the uv-workspace member, a linked-worktree `.git` *file*, a nested repository — all correct), F7 (claim withdrawn, test narrowed to match), F9/C8 (`projectconfig` imports with no Sphinx at all; `getattr(mod, "setup")` serves the lazy `setup` on all three Sphinxes and under `-M`; a missing sphinx-needs is a clean `ExtensionError`; sibling `sphinxcontrib.*` packages unaffected), F10 (each of the two named mutations fails exactly one of the three new cases), F11, F13 (54 is right), F5.1/F5.2.

**One thing the fix summary does not say, and it is the only thing between this and a plain approve:** dropping `pyproject.toml` as a boundary — necessary for the workspace layout — also removed the only thing that stopped the walk in a tree with **no `.git`**: an unpacked sdist, a CI artefact directory, an exported docs tree. Measured with the same directories on both heads: this head adopts a stranger's `ubproject.toml` two levels above the sdist root, where `66900479` stopped at the sdist. It is *announced* (the `Applied … from <path>` line is `info` and prints without `-v`), so it is not silent, but the reach is wider than before. A three-line close that keeps every layout this PR added working: after the loop falls off the end without ever seeing `.git`, either fall back to `pyproject.toml` as the boundary on a second pass, or say in the `Applied` line that the file lies outside any repository root. Plus the sdist layout in `TestDiscovery`.

Small things worth folding into the same push: `ExtensionError(f"Could not import extension {__name__}", error)` — Sphinx appends `(exception: …)` itself, so it currently prints twice; `test_reports.py:114` still says "the convert CLI"; one subprocess test for the missing-sphinx-needs message would pin the shape that the lazy `setup` now owns. Nothing else remaining is new: F4 and the ubCode `[test_reports]` model are deferred as agreed; F6, F8, F12, F14 are follow-ups.

With the sdist boundary handled, I'm happy for this to merge.

docs/configuration.rst
```diff
@@ -401,3 +403,102 @@ An example of a JSON file, which supports the below configuration, can be seen i
...
+.. code-block:: toml
+
+   [test_reports]
+   file_option = "report_file"
+   source_file_option = "file"
+   source_line_option = "line"
+   deterministic_case_ids = true
+   extra_options = ["more_info", "priority"]
+   property_link_types = { request = "req" }
+   rootdir = "docs"
```
MaximilianSoerenPollak, line 432: What des the `root dir` here actually denote? 
Root_dir of what? Where the test_cases will be created?

sphinxcontrib/test_reports/projectconfig.py
```diff
@@ -0,0 +1,436 @@
...
+        return "an integer"
+    if expected is bool:
+        return "a boolean"
+    if expected is str:
+        return "a string"
+    if expected is list:
+        return "an array of strings"
+    if expected is dict:
+        return "a table"
+    return expected.__name__
```
MaximilianSoerenPollak, line 389: Nit (if this is meant mutual exclusive): 


Then with `if => if => if` it will happen that once one condition is met it will still `check` all others. 
Therefore and ìfelse` might be better if these  options are 'mutually exclusive'

tests/doc_test/ubproject_toml/conf.py
```diff
@@ -0,0 +1,23 @@
...
+master_doc = "index"
+
+project = "ubproject-toml-test"
+copyright = "2026, test"
+author = "test"
+version = "1.0"
+release = "1.0"
+language = "en"
+exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
+html_theme = "alabaster"
```
MaximilianSoerenPollak, line 18: Q: 

Could other themes potentially break this rendering of the testcases etc.? Or is it not tightly couple to it?

MaximilianSoerenPollak (dismissed):
For me, as far as I can juddge from my knowledge, this looks fine. 

I had two/three comments but they are nothing that blocks Approvment from my side. 

However the Comment from Chris probably should be addressed as it seems a genuine concern / issue.

docs/configuration.rst
ubmarco: Fair question, and the docs didn't answer it. `rootdir` is the directory that the *relative report paths in the directives* are resolved against: with `rootdir = "docs"`, `.. test-file:: reports/pytest.xml` reads `docs/reports/pytest.xml`. It is an input location — nothing is written there, and no needs are created in it.

Documented in d376c93: a comment on the example line, and the **Paths** paragraph now says it with a worked path and links to the `tr_rootdir` section. That section had no label to link to (nor did `tr_report_template`, the only other one missing), so those are added too.

sphinxcontrib/test_reports/projectconfig.py
ubmarco: They are mutually exclusive, but each branch `return`s, so nothing after the first match is evaluated — `elif` would be identical in behaviour here, only in style. Leaving it as the flat `condition → label` form, since that reads as the lookup table it is. Happy to switch if you'd still rather see `elif`.

tests/doc_test/ubproject_toml/conf.py
ubmarco: Not coupled. The theme only styles the surrounding page; test-case and test-file rendering comes from sphinx-needs' own layouts and this extension's `tr_report_template`, and this fixture's tests assert config values and need data rather than rendered HTML. `alabaster` is here only because every other fixture uses it — 25 of the 26 `tests/doc_test/*/conf.py` set a theme, and all 25 set `alabaster` — so the fixture is consistent with its neighbours rather than picking anything meaningful.

ubmarco:
Sdist boundary closed in d376c93, along with the three small ones. Re [round 2](https://github.com/useblocks/sphinx-test-reports/pull/145#pullrequestreview-5130895055).

**The boundary.** I took the first of your two options, but not in the shape you sketched — a post-loop fallback would never have fired. In the layout you measured the loop does not fall off the end: it *succeeds*, returning the stranger's `ubproject.toml` two levels up, so there is no "after the loop" to fall back from. The boundary has to be known before the walk starts.

So it is computed first: the repository root when a `.git` exists anywhere above the start, and only otherwise the distribution root (`pyproject.toml`). Inside a repository a `pyproject.toml` on the way up still bounds nothing, so `docs/pyproject.toml` beside `conf.py` and the uv-workspace member are untouched — the second boundary only exists where the first one does not. A file sitting *on* either boundary is still found, so an sdist shipping its own `ubproject.toml` is configured by it, and the fruitless-search report now names which of the two roots ended the walk.

Verified the way you framed it: the loader test went red first on your exact layout (it returned the stranger's path), then green; the same layout as a real build no longer adopts it and logs `no ubproject.toml in <docs> or its parents up to the distribution root <sdist> (holding pyproject.toml)`. Four `TestDiscovery` cases: the sdist, the file on the distribution root, a repository marker outranking a distribution root, and the report. 169/169, ruff, strict mypy and a `-W` docs build green on the floor cell (3.11.15 / sphinx 7.4.7 / needs 6.0.1) as well as on 3.14 with current sphinx and needs.

**What is still open, deliberately:** a tree with *neither* marker — a bare exported docs tree, some CI artefact directories — has nothing to bound it and still reaches the filesystem root. Your second option covers that residue, and I did not take it: say the word and the `Applied … from <path>` line grows the "outside any repository root" clause. Right now it prints the absolute path and nothing more.

**The three small ones.** The `ExtensionError` printed the exception twice — the message spelled out `(exception: …)` *and* passed `orig_exc`, which `__str__` renders itself; the message is now bare and the wrapped exception is the second argument only. A subprocess test blocks `sphinx_needs`, catches the error and asserts `(exception:` appears exactly once, which also pins the shape the lazy `setup` owns. `test_reports.py:114` says `test-reports build needs`.

Maximilian's three are answered in-thread; the `rootdir` one was a real docs gap and is fixed in the same push.

chrisjsewell (commented):
Round 3, on `d376c935`. You were right that a post-loop fallback could never have fired — the walk *succeeds* on the stranger's file — and computing the boundary before the walk is the correct shape. Verified by construction against the loader, seven layouts:

| layout | result |
|---|---|
| sdist under a stranger's `ubproject.toml` two levels up | not adopted; reports `up to the distribution root … (holding pyproject.toml)` |
| sdist with its own `ubproject.toml` on its root | found |
| repository, `docs/pyproject.toml` beside `conf.py` | the root file is reached |
| repository, uv-workspace member `packages/x/docs` | the root file is reached |
| repository without the file, a stranger's above the repository | not adopted; `up to the repository root … (holding .git)` |
| linked worktree (`.git` is a file), stranger above | not adopted — the repository marker still wins |
| neither marker anywhere | walks up and adopts — the residue you named |

169/169 (with the venv first on `PATH` — the three `sphinx-build` subprocess tests pick up whatever `sphinx-build` is on `PATH`, which is worth remembering for anyone running the suite from an unactivated shell), `ruff check`, `ruff format --check`, strict `mypy` all clean. The `ExtensionError` now prints `(exception: …)` once and the subprocess test pins that; the `convert CLI` comment is gone.

On the residue: yes, take the second option too — the `Applied … from <path>` line saying the file lies outside any repository or distribution root costs a clause and turns the last silent-wrong-config route into an announced one. Not a blocker; it can follow.

Approving.

chrisjsewell (approved):

PR #147, opened 2026-09-05
build: require Python 3.11

ubmarco:
Python 3.10 reached the end of upstream support.

The floor matters beyond housekeeping: `tomllib` landed in 3.11, so a 3.10 floor means carrying `tomli` as a dependency purely for older interpreters. Split out of #145, which needs a stdlib TOML parser, so the version bump can be reviewed and merged on its own.

## Changes

- `requires-python` `>=3.10` → `>=3.11`
- `noxfile.py` and the CI test matrix run 3.11/3.12
- the `3.9` and `3.10` classifiers go (3.9 was already stale — `requires-python` has excluded it for some time)
- the `pre-commit` and `linkcheck` jobs pinned Python 3.9 purely to drive nox; they move to 3.12, since 3.9 is two releases past end of life and the sessions they start pick their own interpreter anyway

No source change — nothing in the package uses a 3.11 feature yet. 110 tests pass.

For reference, the sibling repos are already ahead of this: sphinx-codelinks and sphinx-mounts both require `>=3.12`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.10", "3.12"] # No "3.10", as nose seem to have problems with it
+        python-version: ["3.11", "3.12"]
         sphinx-version: ["5.0", "8.1.3"]
         sphinx_needs-version: ["2.1", "4.2", "5.1", "6.3.0", "8.0.0"]
     steps:
@@ -30,7 +30,7 @@ jobs:
       - name: Set Up Python
         uses: actions/setup-python@v2
         with:
-          python-version: "3.9"
+          python-version: "3.12"
       - name: Install Nox Dependencies
         run: |
           python -m pip install poetry nox nox-poetry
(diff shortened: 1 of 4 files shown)


chrisjsewell (commented):
Read the diff and checked the tree at `8bde9fda` for anything the diff cannot show. It is correct and complete as far as I can find:

- Nothing else in the repository names 3.10 or 3.9 as a floor: no `ruff` `target-version`, no `mypy` `python_version`, no docs sentence — so `ruff` now targets 3.11 through `requires-python` alone, which is the right way round.
- The removed matrix comment was already false as written (it listed `"3.10"` while saying there was none), so nothing is lost with it.
- CI: 26 checks green on this head.

One nit in the description: the last sentence is stale. Since useblocks/sphinx-needs#1871 both sphinx-codelinks and sphinx-mounts require `>=3.11`, which is the workspace's floor — so 3.11 here is not "behind" the siblings, it is exactly aligned with them.

One FYI for the import, not for this PR: the workspace's manifest check wants `requires-python` to be exactly `>=3.11,<4` (the root's string). Adding the `,<4` is harmless whenever it happens and can equally wait for the import PR.

I would approve this as it stands.

MaximilianSoerenPollak (approved):
For me this seems fine. 
With the comment above seems like all other things inside the codebase have been taken care of as well.

PR #148, opened 2026-09-06
feat(cli): `test-reports build needs` — Sphinx-free test-XML to needs.json, configured from [test_reports.build.needs]

ubmarco:
Based directly on master, independent of useblocks/sphinx-test-reports#151 (pytest plugin). Both touch `docs/changelog.rst` and `docs/index.rst`, so whichever lands second needs a trivial rebase.

Restores the converter from useblocks/sphinx-test-reports#144 — merged ahead of the configuration work and reverted in useblocks/sphinx-test-reports#146 — rebuilt on top of the declarative config so it reads its settings from `ubproject.toml` from the start.

## What

`test-reports build needs FILE... --output needs.json` (named as `ubc build needs` is) turns test-result XML into a schema-conform `needs.json` **without running Sphinx**, so a build system can schedule and cache the conversion and the documentation build only imports the result.

- **No Sphinx import** anywhere in the command's import chain (asserted by a test; the lazy `setup` it relies on arrived with useblocks/sphinx-test-reports#145).
- **Byte-stable output**: sorted keys, fixed indent, no wall clock.
- **Shaped like the build's own needs.** A converted need has exactly the fields the `test-case` directive creates, with the same values in them — the title is the case name, `result` is spelled as the build spells it (`passed`, `failure`, `error`, `skipped`, `disabled`), so one filter matches imported and local cases alike — `suite`, `case`, `case_name`, `case_parameter`, `classname`, `result`, `time`, the report path under `file`, the source location under `case_file`/`case_line` — with the three renameable field names taken from the same `[test_reports]` section the build reads (`file_option`, `source_file_option`, `source_line_option`; defaults defined once, in `projectconfig.DEFAULT_FIELD_NAMES`). XML properties become fields only when listed in `extra_options` — the list that makes the build register and accept them — or added with `--extra-option` (additive; a name the file's list lacks is reported, because the build drops it on import). A listed property is written for every case, `null` where a case has none; the rest is reported once. An end-to-end test `needimport`s a converted file into a real build driven by the same `ubproject.toml` and asserts nothing is dropped.
- Deterministic case IDs, source location, synthesized source URLs (`https://`, `ssh://[:port]`, `git://` and scp-style remotes); `--link-property` promotes properties to link fields; a report given twice, or any case reported twice, is refused rather than silently collapsed. Credentials in the remote URL are stripped before it is written into the needs, and a report without a single test case is reported instead of quietly adding nothing.

## Configuration: `[test_reports.build.needs]`

The command's settings live in a table named after the command — `build` holds one table per artifact it produces — inside the shared `[test_reports]` section of `ubproject.toml` — the file the Sphinx build reads (#145) — so the two consumers of a project never work from different descriptions of it.

```toml
[test_reports]
extra_options = ["TestType", "DerivationTechnique"]

[test_reports.build.needs]
project = "My Project"
version = "2.0"
need_type = "testcase"
tags = ["ci", "unit"]
link_properties = { PartiallyVerifies = "partially_verifies" }
remote_url = "https://github.com/org/repo"
url_pattern = "{base}/blob/{commit}/{file}#L{line}"
```

- **Precedence**: flag > table > built-in default. Flags stay for per-invocation values such as `--commit "$(git rev-parse HEAD)"`.
- **Lookup** mirrors the build: `ubproject.toml` is searched for upwards from the working directory to the project root. `--config PATH` names a file explicitly; `--no-config` makes the output depend only on the arguments. The file the search finds is named on stderr on every run, since it may sit directories above and the output depends on it.
- **Validation** follows the section's policy (#145): a known key with the wrong type is fatal, an unknown key is reported and dropped, table values are checked. The **whole** section is validated, and the build validates the `build.needs` table too — one file, one verdict. `url_pattern` is checked before any report is read. A renameable field may not take a fixed field's name (`case`, `result`, ...), in the file or in `conf.py`; the renames must live in `ubproject.toml` for the converter to see them.
- **Cross-key rule**: `build.needs.need_type` and the `type` of `case` both name the need type of a test case; a side that is not set is compared at its default, and the command re-checks the *merged* value so a `--need-type` flag cannot bypass the rule.

## Review

Reviewed in this PR (comment below): 30 verified findings from the correctness pass plus a manual pass, all addressed. The CLI-side findings of the useblocks/sphinx-test-reports#145 review land here as well. Importing a converted file needs sphinx-needs ≥ 4 (older `needimport` read `description`, not `content`) — moot with useblocks/sphinx-test-reports#150's floor.

Second round — @chrisjsewell's review (five blocking, eight worth fixing, the nits) and @MaximilianSoerenPollak's questions — addressed at `d8940fc` in six commits (`423e51b..d8940fc`), each green on its own.

## Tests

162 on top of master (331 in total): `test_cli_convert.py` (conversion, `needimport` into a real build), `test_cli_config.py` (precedence, lookup, diagnostics), `test_needs_export.py` (writer and URL helpers), `test_identity.py` (the `name[param]` split), and the field-name and table validation in `test_project_config.py`.

### DIFF
--- a/sphinxcontrib/test_reports/cli.py
+++ b/sphinxcontrib/test_reports/cli.py
@@ -1,0 +1,184 @@
+"""``test-reports`` command line interface.
+
+Converts test-result XML into needs.json *outside* Sphinx, so a build system can
+schedule and cache the conversion and the documentation build only imports the
+result. Nothing in the import chain of this module may import Sphinx; the test
+suite asserts that.
+"""
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+from sphinxcontrib.test_reports.junitparser import JUnitParser
+from sphinxcontrib.test_reports.needs_export import DEFAULT_VERSION, build_needs_file
+from sphinxcontrib.test_reports.remote import DEFAULT_URL_PATTERN, normalise_remote_url
+
+
+def _build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="test-reports",
+        description="Convert test-result XML into sphinx-needs data.",
+    )
+    subcommands = parser.add_subparsers(dest="command", required=True)
+
+    convert = subcommands.add_parser(
+        "convert",
+        help="Convert one or more test-result XML files into needs.json.",
+        description=(
+            "Convert one or more test-result XML files into a needs.json that "
+            "can be consumed with needimport or needs_external_needs."
+        ),
+    )
+    convert.add_argument(
+        "files",
+        nargs="+",
+        metavar="FILE",
+        help="Test-result XML files (a build system passes these as a file list).",
+    )
+    convert.add_argument(
+        "--output",
+        "-o",
+        required=True,
+        metavar="PATH",
+        help="Where to write the needs.json.",
+    )
+    convert.add_argument(
+        "--project",
+        default="",
+        help="Project name recorded in the needs.json envelope.",
+    )
+    convert.add_argument(
+        "--version",
+        default=DEFAULT_VERSION,
+        help=f"Version key in the envelope (default: {DEFAULT_VERSION}).",
+    )
+    convert.add_argument(
+        "--need-type",
+        default="testcase",
+        help="Need type for each test case (default: testcase).",
+    )
+    convert.add_argument(
+        "--tags",
+        default="",
+        help="Comma-separated tags applied to every created need.",
+    )
+    convert.add_argument(
+        "--link-property",
+        action="append",
+        default=[],
+        metavar="PROPERTY=LINK_FIELD",
+        help=(
+            "Promote an XML property to a link field, comma-splitting its value "
+            "(repeatable), e.g. PartiallyVerifies=partially_verifies."
+        ),
+    )
+    convert.add_argument(
+        "--remote-url",
+        default="",
+        help="Repository URL used to synthesize source links; git remotes are accepted.",
+    )
+    convert.add_argument(
+        "--commit",
+        default="",
+        help="Commit-ish the reports were produced from.",
+    )
+    convert.add_argument(
+        "--url-pattern",
+        default=DEFAULT_URL_PATTERN,
+        help=f"Source-URL template (default: {DEFAULT_URL_PATTERN}).",
+    )
+    return parser
+
+
+def _parse_link_properties(values: list[str]) -> dict[str, str]:
+    mapping = {}
+    for value in values:
+        property_name, separator, link_field = value.partition("=")
+        if not separator or not property_name.strip() or not link_field.strip():
+            raise ValueError(
+                f"--link-property expects PROPERTY=LINK_FIELD, got {value!r}"
+            )
+        mapping[property_name.strip()] = link_field.strip()
+    return mapping
+
+
+def _warn_about_absent_source_lines(path: Path, suites: list) -> None:
+    """Report the most common cause of a missing source location.
+
+    pytest emits ``file``/``line`` as ``<testcase>`` attributes only under
+    ``junit_family = xunit1`` (or ``legacy``); its default ``xunit2`` filters
+    them out, which silently costs the source location of every case.
+    """
+    cases = [case for suite in suites for case in suite.get("testcases", [])]
+    if cases and all(case.get("line", -1) == -1 for case in cases):
+        print(
+            f"warning: {path}: no <testcase> carries a 'line' attribute, so no "
+            "source location could be recorded. pytest emits file/line only "
+            "with junit_family = xunit1 (or legacy); its default xunit2 drops "
+            "them.",
+            file=sys.stderr,
+        )
+
+
+def _convert(arguments: argparse.Namespace) -> int:
+    try:
+        link_properties = _parse_link_properties(list(arguments.link_property))
+    except ValueError as error:
+        print(f"error: {error}", file=sys.stderr)
+        return 2
+
+    if bool(arguments.remote_url) != bool(arguments.commit):
+        print(
+            "error: --remote-url and --commit must be given together; without "
+            "both, no source URL can be synthesized.",
+            file=sys.stderr,
+        )
+        return 2
+
+    suites: list = []
+    for name in arguments.files:
+        path = Path(name)
+        if not path.is_file():
+            print(f"error: no such file: {path}", file=sys.stderr)
+            return 1
+        try:
+            parsed = JUnitParser(str(path)).parse()
+        except Exception as error:  # noqa: BLE001 - report, never traceback
+            print(f"error: {path}: {error}", file=sys.stderr)
+            return 1
+        _warn_about_absent_source_lines(path, parsed)
+        suites.extend(parsed)
+
+    payload = build_needs_file(
+        suites,
+        project=arguments.project,
+        version=arguments.version,
+        need_type=arguments.need_type,
+        tags=[tag.strip() for tag in arguments.tags.split(",") if tag.strip()],
+        link_properties=link_properties,
+        base_url=normalise_remote_url(arguments.remote_url),
+        commit=arguments.commit,
+        url_pattern=arguments.url_pattern,
+    )
+
+    output = Path(arguments.output)
+    output.parent.mkdir(parents=True, exist_ok=True)
+    # Sorted keys and a fixed indent keep the output byte-stable across runs.
+    output.write_text(
+        json.dumps(payload, indent=2, sort_keys=True) + "\n", encoding="utf-8"
+    )
+    return 0
+
+
+def main(argv: "list[str] | None" = None) -> int:
+    """Entry point. Returns a process exit code instead of raising."""
+    arguments = _build_parser().parse_args(argv)
+    if arguments.command == "convert":
+        return _convert(arguments)
+    return 2  # pragma: no cover - argparse rejects unknown commands
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/sphinxcontrib/test_reports/cli.py
+++ b/sphinxcontrib/test_reports/cli.py
@@ -26,6 +26,7 @@ from sphinxcontrib.test_reports.projectconfig import (
     DEFAULT_TOML_FILENAME,
     SECTION,
     TomlConfigError,
+    case_need_type,
     find_project_config,
     load_project_config,
 )
@@ -33,6 +34,10 @@ from sphinxcontrib.test_reports.remote import DEFAULT_URL_PATTERN, normalise_rem
 
 #: How the table is spelled in help texts and diagnostics.
 TABLE = f"[{SECTION}.{CONVERT_TABLE}]"
+
+
+def _warn(message: str) -> None:
+    print(f"warning: {message}", file=sys.stderr)
 
 
 def _build_parser() -> argparse.ArgumentParser:
@@ -191,14 +196,18 @@ def _warn_about_absent_source_lines(path: Path, suites: list) -> None:
         )
 
 
-def _load_convert_table(
+def _load_section(
     arguments: argparse.Namespace,
 ) -> "tuple[dict, Path | None, str | None]":
-    """Load ``[test_reports.convert]``, honouring ``--config``/``--no-config``.
+    """Load the ``[test_reports]`` section, honouring ``--config``/``--no-config``.
 
-    Returns ``(table, path, error_message)``. ``table`` is ``{}`` and ``path``
-    ``None`` when no file applies: an absent *default* file is not an error, an
-    explicitly given path that cannot be read is.
+    Returns ``(section, path, error_message)``. ``section`` is ``{}`` and
+    ``path`` ``None`` when no file applies: an absent *default* file is not an
+    error, an explicitly given path that cannot be read is.
+
+    The whole section is loaded, not only the converter's table: validation
+    covers the file as the build sees it, and the build's ``case`` entry is
+    needed to check the need type against.
 
     The default file is searched for upwards from the working directory, since
     it conventionally sits at the project root while the converter runs from
@@ -218,15 +227,11 @@ def _load_convert_table(
             return {}, None, None
         path = found
 
-    def warn(message: str) -> None:
-        print(f"warning: {message}", file=sys.stderr)
-
     try:
-        section = load_project_config(path, warn) or {}
+        section = load_project_config(path, _warn) or {}
     except TomlConfigError as error:
         return {}, path, f"error: {error}"
-    table = section.get(CONVERT_TABLE, {})
-    return table if isinstance(table, dict) else {}, path, None
+    return section, path, None
 
 
 #: Built-in value of every conversion setting, used when neither a flag nor
@@ -243,7 +248,8 @@ _DEFAULTS: "dict[str, object]" = {
     "commit": "",
     "url_pattern": DEFAULT_URL_PATTERN,
 }
-assert set(_DEFAULTS) == set(CONVERSION_KEYS), "every conversion key needs a default"
+if set(_DEFAULTS) != set(CONVERSION_KEYS):  # pragma: no cover - import-time guard
+    raise RuntimeError("every [test_reports.convert] key needs a built-in default")
 
 #: argparse destination per conversion key, where it differs from the key.
 #: Only the repeatable ``--link-property`` flag does.
@@ -285,8 +291,13 @@ def _resolve_settings(
 
 
 def _spell(key: str, sources: "dict[str, str]") -> str:
-    """The key as the user wrote it: the flag, or the TOML key."""
-    return key if sources.get(key) == "toml" else _FLAGS[key]
+    """The key as the user wrote it: the flag, the TOML key, or the default."""
+    source = sources.get(key)
+    if source == "toml":
+        return key
+    if source == "default":
+        return f"the default {key}"
+    return _FLAGS[key]
 
 
 def _pair_requirement(sources: "dict[str, str]", path: "Path | None") -> str:
@@ -295,19 +306,44 @@ def _pair_requirement(sources: "dict[str, str]", path: "Path | None") -> str:
     Either half can come from the TOML file, so naming only the flags would
     point at options that appear nowhere in the invocation.
     """
-    requirement = f"{_spell('remote_url', sources)} and {_spell('commit', sources)}"
+    # The half that is missing is named by the flag that would supply it --
+    # that is the actionable spelling -- and the half that was given by however
+    # the user gave it.
+    names = [
+        key if sources.get(key) == "toml" else _FLAGS[key]
+        for key in ("remote_url", "commit")
+    ]
+    requirement = f"{names[0]} and {names[1]}"
     if "toml" in (sources.get("remote_url"), sources.get("commit")):
         requirement += f" (the bare names are {TABLE} keys in {path})"
     return requirement
 
 
 def _convert(arguments: argparse.Namespace) -> int:
-    table, config_path, error = _load_convert_table(arguments)
+    section, config_path, error = _load_section(arguments)
     if error:
         print(error, file=sys.stderr)
         return 2
 
-    settings, sources = _resolve_settings(arguments, table)
+    table = section.get(CONVERT_TABLE, {})
+    settings, sources = _resolve_settings(
+        arguments, table if isinstance(table, dict) else {}
+    )
+
+    # The loader already rejects a file whose convert.need_type disagrees with
+    # case's type. A --need-type flag (or the default) is not in the file, so
+    # the merged value has to be checked here as well, or the converter writes
+    # needs of a type the build does not register.
+    case_type = case_need_type(section)
+    if case_type is not None and settings["need_type"] != case_type:
+        print(
+            f"error: {_spell('need_type', sources)} is {settings['need_type']!r} "
+            f"but [{SECTION}] case's type is {case_type!r} in {config_path}. Both "
+            f"name the need type of a test case -- need_type for this converter, "
+            f"case for the Sphinx build -- so they must agree.",
+            file=sys.stderr,
+        )
+        return 2
 
     try:
         link_properties = _parse_link_properties(settings["link_properties"])
@@ -337,17 +373,22 @@ def _convert(arguments: argparse.Namespace) -> int:
         _warn_about_absent_source_lines(path, parsed)
         suites.extend(parsed)
 
-    payload = build_needs_file(
-        suites,
-        project=settings["project"],
-        version=settings["version"],
-        need_type=settings["need_type"],
-        tags=settings["tags"],
-        link_properties=link_properties,
-        base_url=normalise_remote_url(settings["remote_url"]),
-        commit=settings["commit"],
-        url_pattern=settings["url_pattern"],
-    )
+    try:
+        payload = build_needs_file(
+            suites,
+            project=settings["project"],
+            version=settings["version"],
+            need_type=settings["need_type"],
+            tags=settings["tags"],
+            link_properties=link_properties,
+            base_url=normalise_remote_url(settings["remote_url"]),
+            commit=settings["commit"],
+            url_pattern=settings["url_pattern"],
+            warn=_warn,
+        )
+    except ValueError as error:  # duplicate test cases across the inputs
+        print(f"error: {error}", file=sys.stderr)
+        return 2
 
     output = Path(arguments.output)
     output.parent.mkdir(parents=True, exist_ok=True)
--- a/sphinxcontrib/test_reports/cli.py
+++ b/sphinxcontrib/test_reports/cli.py
@@ -19,7 +19,12 @@ import sys
 from pathlib import Path
 
 from sphinxcontrib.test_reports.junitparser import JUnitParser
-from sphinxcontrib.test_reports.needs_export import DEFAULT_VERSION, build_needs_file
+from sphinxcontrib.test_reports.needs_export import (
+    DEFAULT_VERSION,
+    build_needs_file,
+    iter_cases,
+    optional,
+)
 from sphinxcontrib.test_reports.projectconfig import (
     CONVERSION_KEYS,
     CONVERT_TABLE,
@@ -27,10 +32,15 @@ from sphinxcontrib.test_reports.projectconfig import (
     SECTION,
     TomlConfigError,
     case_need_type,
+    field_names,
     find_project_config,
     load_project_config,
 )
-from sphinxcontrib.test_reports.remote import DEFAULT_URL_PATTERN, normalise_remote_url
+from sphinxcontrib.test_reports.remote import (
+    DEFAULT_URL_PATTERN,
+    check_url_pattern,
+    normalise_remote_url,
+)
 
 #: How the table is spelled in help texts and diagnostics.
 TABLE = f"[{SECTION}.{CONVERT_TABLE}]"
@@ -103,6 +113,18 @@ def _build_parser() -> argparse.ArgumentParser:
             "Promote an XML property to a link field, comma-splitting its value "
             "(repeatable), e.g. PartiallyVerifies=partially_verifies. Given at "
             "all, it replaces the file's link_properties table."
+        ),
+    )
+    convert.add_argument(
+        "--extra-option",
+        action="append",
+        default=None,
+        metavar="NAME",
+        help=(
+            "Export the XML property NAME as a need field (repeatable). Default: "
+            f"the extra_options of [{SECTION}] -- the same list that makes the "
+            "build accept the field. Properties named by neither are reported "
+            "and left out."
         ),
     )
     convert.add_argument(
@@ -185,8 +207,11 @@ def _warn_about_absent_source_lines(path: Path, suites: list) -> None:
     ``junit_family = xunit1`` (or ``legacy``); its default ``xunit2`` filters
     them out, which silently costs the source location of every case.
     """
-    cases = [case for suite in suites for case in suite.get("testcases", [])]
-    if cases and all(case.get("line", -1) == -1 for case in cases):
+    # Walk the report the way the export does: the parser files the cases of
+    # a suite with nested <testsuite> elements under testsuite_nested only, so
+    # looking at the top-level testcases alone goes quiet on nested reports.
+    cases = [case for _suite_name, case in iter_cases(suites)]
+    if cases and all(optional(case.get("line"), -1) == "" for case in cases):
         print(
             f"warning: {path}: no <testcase> carries a 'line' attribute, so no "
             "source location could be recorded. pytest emits file/line only "
@@ -319,6 +344,14 @@ def _pair_requirement(sources: "dict[str, str]", path: "Path | None") -> str:
     return requirement
 
 
+def _extra_options(arguments: argparse.Namespace, section: dict) -> list:
+    """Properties exported as fields: the flag if given, else the section's."""
+    if arguments.extra_option is not None:
+        return list(arguments.extra_option)
+    names = section.get("extra_options", [])
+    return list(names) if isinstance(names, list) else []
+
+
 def _convert(arguments: argparse.Namespace) -> int:
     section, config_path, error = _load_section(arguments)
     if error:
@@ -351,6 +384,17 @@ def _convert(arguments: argparse.Namespace) -> int:
         print(f"error: {error}", file=sys.stderr)
         return 2
 
+    problem = check_url_pattern(str(settings["url_pattern"]))
+    if problem is not None:
+        # Checked before any report is read: str.format would otherwise fail
+        # on the first case that carries a file, as a traceback.
+        print(
+            f"error: {_spell('url_pattern', sources)} {settings['url_pattern']!r}: "
+            f"{problem}",
+            file=sys.stderr,
+        )
+        return 2
+
     if bool(settings["remote_url"]) != bool(settings["commit"]):
         print(
             f"error: {_pair_requirement(sources, config_path)} must be given "
@@ -359,7 +403,7 @@ def _convert(arguments: argparse.Namespace) -> int:
         )
         return 2
 
-    suites: list = []
+    reports: list = []
     for name in arguments.files:
         path = Path(name)
         if not path.is_file():
@@ -371,11 +415,13 @@ def _convert(arguments: argparse.Namespace) -> int:
             print(f"error: {path}: {error}", file=sys.stderr)
             return 1
         _warn_about_absent_source_lines(path, parsed)
-        suites.extend(parsed)
+        # The path as given, like the build records the path given to its
+        # directives; it is a label for the report, not something resolved.
+        reports.append((str(path), parsed))
 
     try:
         payload = build_needs_file(
-            suites,
+            reports,
             project=settings["project"],
             version=settings["version"],
             need_type=settings["need_type"],
@@ -384,6 +430,11 @@ def _convert(arguments: argparse.Namespace) -> int:
             base_url=normalise_remote_url(settings["remote_url"]),
             commit=settings["commit"],
             url_pattern=settings["url_pattern"],
+            # The renameable field names and the accepted extra fields come
+            # from the same section the build reads, so an imported need has
+            # the shape of a local one.
+            fields=field_names(section),
+            extra_options=_extra_options(arguments, section),
             warn=_warn,
         )
     except ValueError as error:  # duplicate test cases across the inputs
--- a/sphinxcontrib/test_reports/cli.py
+++ b/sphinxcontrib/test_reports/cli.py
@@ -1,6 +1,6 @@
 """``test-reports`` command line interface.
 
-Converts test-result XML into needs.json *outside* Sphinx, so a build system can
+``test-reports build needs`` turns test-result XML into needs.json *outside* Sphinx, so a build system can
 schedule and cache the conversion and the documentation build only imports the
 result. Nothing in the import chain of this module may import Sphinx; the test
 suite asserts that.
@@ -57,12 +57,20 @@ def _build_parser() -> argparse.ArgumentParser:
     )
     subcommands = parser.add_subparsers(dest="command", required=True)
 
-    convert = subcommands.add_parser(
-        "convert",
-        help="Convert one or more test-result XML files into needs.json.",
+    # `build <artifact>`, as ubCode spells it (`ubc build needs`): the command
+    # names what gets produced, and leaves room for further artifacts.
+    build = subcommands.add_parser(
+        "build",
+        help="Build an artifact from test-result XML.",
+        description="Build an artifact from one or more test-result XML files.",
+    )
+    artifacts = build.add_subparsers(dest="artifact", required=True)
+    convert = artifacts.add_parser(
+        "needs",
+        help="Build a needs.json from one or more test-result XML files.",
         description=(
-            "Convert one or more test-result XML files into a needs.json that "
-            "can be consumed with needimport or needs_external_needs. Settings "
+            "Build a needs.json from one or more test-result XML files, ready "
+            "to be consumed with needimport or needs_external_needs. Settings "
             f"default to the {TABLE} table of {DEFAULT_TOML_FILENAME}; a flag "
             "overrides the file's value for that key."
         ),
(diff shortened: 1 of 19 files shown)


ubmarco:
## Review

Reviewed the restored converter at `91e71b8` (manual pass plus the correctness engine: 4 finder angles → 30 candidates → 23 verifiers; 30 verified, 1 refuted). Everything below is addressed in `9b6a695` and `0407f43`; the fix for each item is linked at its final location in a8c7370699e721aa2848931a41e4bb7ad3b7fb57.

### 1. [High] The documented `needimport` path dropped or misfiled fields

The converter wrote `file`/`line`/`name`/`result_text`/`remote_url` and every XML property, while the build registers `case_file`/`case_line`/`case_name` and only the properties in `tr_extra_options`, and defines `file` as the XML *report* path. Importing a converted file warned `Unknown keys in import need source: ['line', 'name', 'remote_url', 'result_text', ...]` and deleted them; the source path landed in the report-path field. Two consumers of one project described the same test case differently.

**Fixed:** a converted need now has exactly the shape of the need the `test-case` directive creates, with the three renameable field names read from the same `[test_reports]` section (defaults defined once, in `DEFAULT_FIELD_NAMES`, and used by the extension's config values):

https://github.com/useblocks/sphinx-test-reports/blob/a8c7370699e721aa2848931a41e4bb7ad3b7fb57/sphinxcontrib/test_reports/needs_export.py#L214-L220

Properties become fields only when listed in `extra_options` — the list that makes the build accept them — or given with the new `--extra-option`; the rest is reported once with the key to add. `result_text`/`remote_url` are registered by the extension. `identity.split_case_name` replaces the directive's inline regex so both sides split `name[param]` alike. `test_converted_needs_import_without_loss` now converts a report and `needimport`s it into a real build driven by the same `ubproject.toml`, asserting no key is dropped.

### 2. [High] Overlapping reports collapsed silently

The same case in two inputs (same report twice, two platforms, rerun attempts) collapsed into one need by plain dict assignment — a failing result replaced by a passing one, `needs_amount` undercounting, exit 0, nothing on stderr. Reproduced: `convert gtest.xml gtest.xml` → 5 needs, exit 0.

**Fixed:** duplicates are refused with the IDs named; a needs.json cannot hold two needs with one ID, and keeping either silently loses evidence.

https://github.com/useblocks/sphinx-test-reports/blob/a8c7370699e721aa2848931a41e4bb7ad3b7fb57/sphinxcontrib/test_reports/needs_export.py#L309-L313

### 3. [High] The need-type agreement rule had two holes

The loader compared `convert.need_type` with `case.type` only when *both* were set — a customised `case` next to a `convert` table without `need_type` (or the mirror) passed — and a `--need-type` flag is not in the file at all, so it bypassed the rule entirely. Reproduced: file with `case.type = "check"`, `--need-type testcase` → exit 0, wrote `testcase` needs the build never registers.

**Fixed:** a side that is not set is compared at its default (only when the `convert` table exists — a build-only project names its case type freely), and the CLI re-checks the *merged* value:

https://github.com/useblocks/sphinx-test-reports/blob/a8c7370699e721aa2848931a41e4bb7ad3b7fb57/sphinxcontrib/test_reports/cli.py#L368-L374

### 4. [Medium] Evidence text lost its indentation

`_literal_block` stripped every line's leading whitespace before re-indenting, flattening tracebacks and assertion diffs into a column. **Fixed:** the body is dedented as a whole (the common indentation XML pretty-printing adds) and re-indented, so relative indentation survives.

https://github.com/useblocks/sphinx-test-reports/blob/a8c7370699e721aa2848931a41e4bb7ad3b7fb57/sphinxcontrib/test_reports/needs_export.py#L70-L75

### 5. [Medium] `ssh://host:2222/…` and `git://` remotes produced 404 bases

The port was read as the first path segment (`https://host/2222/org/repo`), and `git://github.com/org/repo.git` became `https://git///github.com/org/repo`. **Fixed:** scheme URLs get their own pattern (the port belongs to the transport and is dropped), unknown schemes pass through untouched.

https://github.com/useblocks/sphinx-test-reports/blob/a8c7370699e721aa2848931a41e4bb7ad3b7fb57/sphinxcontrib/test_reports/remote.py#L20-L28

### 6. [Medium] A bad `url_pattern` was a traceback, late

`{ref}` or an unbalanced brace raised `KeyError`/`ValueError` from `str.format` on the first case that carried a file, as a raw traceback. **Fixed:** the template is checked before any report is read; an error names the problem and the allowed placeholders.

https://github.com/useblocks/sphinx-test-reports/blob/a8c7370699e721aa2848931a41e4bb7ad3b7fb57/sphinxcontrib/test_reports/cli.py#L386-L390

### 7. [Medium] A missing sphinx-needs became a raw traceback

The PEP 562 lazy `setup` moved the ImportError out of the `except ImportError` around Sphinx's `import_module`; `getattr(mod, "setup", None)` only tolerates AttributeError, so instead of `Extension error: Could not import extension …` the user got Sphinx's "please open an issue" crash report. **Fixed:** the ImportError is re-raised as that `ExtensionError` when Sphinx is present to receive it.

https://github.com/useblocks/sphinx-test-reports/blob/a8c7370699e721aa2848931a41e4bb7ad3b7fb57/sphinxcontrib/test_reports/__init__.py#L24-L30

### 8. [Medium] The missing-source-location hint was silent for nested suites

It read only the top-level `testcases`, while the parser files the cases of a nested report under `testsuite_nested` — exactly the Ant/Maven shape that most often lacks locations. **Fixed:** it walks the report the way the export does (`iter_cases`); covered with `pytest_nested_example.xml`.

### 9. [Medium] Lowercase IDs vs sphinx-needs' default `needs_id_regex`

`docs/cli.rst` said "just `needimport` it", but the default regex `^[A-Z0-9_]{5,}` refuses every `testcase__…` ID; nothing pointed at the setting. **Fixed** in the docs ("Consuming the result" names the regex to widen, and the link-type requirement for `--link-property` fields).

### 10. [Medium] Build and converter disagreed on `link_properties = { X = "" }`

Emptiness was checked only in the CLI, so the build accepted a file the converter refused. **Fixed:** checked in the loader, once, for both — as is a collision between the two path fields.

### Low

- A property whose name collides with a built-in or link field vanished silently; it is still dropped (overwriting `result` would be worse) but reported through a `warn` callback the CLI routes to stderr.
- `_normalise_convert_table` warns on unknown keys inside `[test_reports.convert]` where #145's pass-through did not. Deliberate: the table is this package's now, so an unknown key there is a typo for this package — same policy, same suppression key (`test_reports.unknown_key`).
- The refuted candidate (quoted forward-reference annotations in `cli.py` on a 3.11 floor) was not acted on.

### Also in this round

`feat(pytest)`: the pytest plugin producing the XML shape the converter consumes, ported from S-CORE docs-as-code's `score_pytest` attribute plugin (see the PR description and `docs/pytest.rst`).

**Tests:** 150 → 264 across the stack; mypy, pre-commit and the docs build are clean locally.

sphinxcontrib/test_reports/cli.py
```diff
@@ -0,0 +1,487 @@
...
+        default=None,
+        metavar="PROPERTY=LINK_FIELD",
+        help=(
+            "Promote an XML property to a link field, comma-splitting its value "
+            "(repeatable), e.g. PartiallyVerifies=partially_verifies. Given at "
+            "all, it replaces the file's link_properties table."
+        ),
+    )
+    needs.add_argument(
+        "--extra-option",
```
ubmarco, line 152: Extra options were renamed to `needs_fields` (extra links are `needs_links`).

ubmarco: This needs a refactoring for `extra_options` and a deprecation. Opened an issue as a follow-up.
https://github.com/useblocks/sphinx-test-reports/issues/157

ubmarco:
Answering the review in https://github.com/useblocks/sphinx-test-reports/pull/148#issuecomment-5562037665.

Re-verified every finding of that review against `a63a20e`, the current head.
All ten plus the Low items are still fixed; the permalinks in the review point
at `a8c7370`, which the later rebases rewrote, so the locations below are the
live ones. Two names in the review have moved since: the table is
`[test_reports.build.needs]`, not `[test_reports.convert]`, and
`_normalise_convert_table` is now `needs_settings`.

| # | Fix, and where it lives now | Covered by |
|---|---|---|
| 1 | A converted need has the shape the `test-case` directive creates: `suite`/`case`/`case_name`/`case_parameter`/`classname`/`result`/`time`, the report path under the `file_option` name and the source location under the `source_*_option` names, all three read from the same section with defaults in `DEFAULT_FIELD_NAMES` — `needs_export.py:214-240`. Properties become fields only when listed in `extra_options` or given with `--extra-option`; the rest is reported once. `identity.split_case_name` splits `name[param]` for both writers. | `test_converted_needs_import_without_loss` (`test_cli_convert.py:514`), which converts and `needimport`s into a real build and asserts no key is dropped |
| 2 | Duplicate cases across inputs are refused with the IDs named, instead of collapsing by dict assignment — `needs_export.py:320-333` | `TestDuplicateCases` (`test_needs_export.py`) |
| 3 | A side that is not set is compared at its default, and the CLI re-checks the *merged* value, so a `--need-type` flag cannot escape the rule — `projectconfig.case_need_type` (`:597`) and `cli.py:420-432` | four cases in `test_cli_config.py:304-379`, plus `test_project_config.py:144` |
| 4 | `_literal_block` dedents the body as a whole and re-indents, so relative indentation survives — `needs_export.py:72-85` | `TestEvidenceText::test_relative_indentation_survives` |
| 5 | Scheme URLs get their own pattern, the port belongs to the transport and is dropped, unknown schemes pass through — `remote.normalise_remote_url` (`:34`) | `TestRemoteUrls`, seven remote spellings |
| 6 | The template is checked before any report is read, naming the problem and the allowed placeholders — `cli.py:441-450` | `TestUrlPatternCheck` |
| 7 | The `ImportError` is re-raised as Sphinx' own `ExtensionError` when Sphinx is there to receive it, so a missing sphinx-needs is a message and not a crash report — `__init__.py:18-34` | — |
| 8 | The missing-source-location hint walks the report with `iter_cases`, so nested Ant/Maven suites are seen — `cli.py:259` | the `pytest_nested_example.xml` case at `test_cli_convert.py:388` |
| 9 | `docs/cli.rst:88-93` names the regex to widen and the link-type requirement for `--link-property` fields | docs build |
| 10 | Empty `PROPERTY = "LINK_FIELD"` names are rejected in the loader, once, for both consumers — `projectconfig.py:504-520` | `test_empty_link_property_names_are_rejected_by_the_loader` (`test_project_config.py:209`), `test_malformed_link_property_value_is_an_error` |

Low items: the colliding property is still dropped and still reported through
the `warn` callback (`needs_export.py:255-262`); unknown keys inside the table
still warn under the same suppression key (`needs_settings`); and the refuted
candidate stands refuted — the quoted forward-reference annotations in `cli.py`
are still there, and today's typing pass added more of them.

Since that review the branch also gained the field declarations in the output
(`needs_schema`, `e4cc0ed`) and lost the two mypy `exclude` entries for the
modules this branch owns (`c08e937`, `a63a20e`). Two findings were deliberately
turned into follow-ups rather than fixed here: `time` as a number (#156) and the
`extra option` → `fields` vocabulary (#157).

State at `a63a20e`: `288 passed`, mypy clean over 11 source files, pre-commit
clean on all files, docs build without warnings.

chrisjsewell (commented):
Reviewed at `a63a20eb` by reproducing the body's claims rather than reading them: three environments (sphinx 9.1 / sphinx-needs 8.5.0; the floor cell 7.4.7 / 6.0.1; and one with no Sphinx installed at all), the full suite on both cells (288 passed each), and nine mutations of the new code against the suite. Verdict: **the design holds and I want this in, after the fixes below.** Five things I'd call blocking, the rest at your discretion.

**What holds, measured.** No `sphinx*` or `docutils*` module is in `sys.modules` after a real conversion in the Sphinx-free environment (lxml is the only import-time need); the lazy `setup` still serves `sphinx-build`, `hasattr` and `dir`, and a `sphinxcontrib` sibling is unaffected. Output is byte-identical across working directories, `TZ` and `PYTHONHASHSEED`, with no timestamp anywhere. A converted need's IDs are string-equal to the ones a real build creates for the same report, and every listed field value matches. Precedence is flag > table > default for every key, lookup follows #145's boundary rule in a repo, a linked worktree, a nested repo and an unpacked tree, `--no-config` makes the output byte-independent of a stray file, and the build refuses the same typo in `build.needs.tags` with the same message. Every remote-URL spelling in the body normalises correctly, duplicate reports and cases are refused with no file written, mypy strict is clean with `test_reports.py` back in, the wheel and sdist carry the entry point and the five modules with no new runtime dependency, and the docs build with zero warnings. Rebased on `7e6a45ae`, `packaging` did not come back.

### Blocking

**1. `--need-type` bypasses the agreement rule when the file has no `[test_reports.case]`.** `cli.py:420` guards the re-check with `if case_type is not None`, the opposite of the loader's rule ("a side that is not set is compared at its default"). All eight combinations measured; seven agree with the body, the eighth does not: a bare `[test_reports]` plus `--need-type check` exits 0 and writes needs of type `check`, and the import then drops every one of them (`Unknown need type 'check'`, once per need, build exit 0 without `-W`). That is the failure the rule exists to prevent, reached through the documented flag. Comparing against `case_need_type(section) or DEFAULT_NEED_TYPE` whenever a config file applied keeps all 288 tests green and turns the reproduction into an error (I tried it).

**2. `--extra-option` replaces the configured list and writes fields the build drops.** With `extra_options = ["TestType"]` in the file, `--extra-option Requirement` prints `warning: properties not exported: PartiallyVerifies, TestType` (the configured one is now dropped) and the build then warns `Unknown keys in import need source: ['Requirement']`. The help text does say "Default: the extra_options of [test_reports]", but a repeatable flag named like that reads as additive, and nothing warns. Either make it additive, or rename it and say so; and since the CLI has the section in hand, warn when an exported name is not in the section's `extra_options`.

**3. The title differs from the directive's.** `cli.rst:38` and `build_need`'s docstring say a converted need is shaped exactly like the one the directive creates and is indistinguishable to a filter or a needtable. Same report, same config: the directive titles the need `Addition`, the converter `MathTest__Addition` (`case_display_name`, the ID digest input). A `needtable` title column, a filter on `title` and every rendered title differ between a built case and the same case imported. Either write `case_display_name` as the title on the directive side too, or drop the claim and document what the title is.

**4. `result` is exported in a different vocabulary.** `RESULT_NAMES = {"failure": "failed"}` is deliberate and `TestResultVocabulary` locks it in, but the user-facing docs never say so. In the mixed project `cli.rst` advertises, `:filter: result == "failure"` matches only the local needs and `result == "failed"` only the imported ones. Map on both sides (the CSS class is computed separately at `test_case.py:156`) or document the rename with the filter consequence.

**5. A project that renames the three fields in `conf.py` only gets silently misfiled data.** With `tr_file_option = "report_file"`, `tr_source_file_option = "file"`, `tr_source_line_option = "line"` and no `ubproject.toml`, the CLI writes `file` (the report path), `case_file`, `case_line`; the build drops the last two with a warning and accepts `file`, which this project defines as the source file, carrying the report path. Your own finding #1, mirrored, and silent on the value that survives. The converter cannot see `conf.py`, so `cli.rst` and `configuration.rst` should say the renames must live in `ubproject.toml`; a stronger fix is a build-side warning when an imported need carries the default names while the configured ones differ.

### Worth fixing

- **Credentials in the remote URL are written into every need.** `https://gitlab-ci-token:$TOKEN@gitlab.example/org/repo.git` normalises with the userinfo intact and lands in `external_url` and `remote_url`, so in a cached artifact and, after import, in published HTML. `--remote-url "$CI_REPOSITORY_URL"` is the natural GitLab spelling and that variable embeds the job token. Strip userinfo in `normalise_remote_url`.
- **A non-report XML converts silently to an empty artifact.** A `pom.xml`, a `coverage.xml` or a genuine report with zero cases gives a valid empty `needs.json`, exit 0, nothing on stderr. For a cached build action "0 needs" is the one outcome that should never be silent.
- **`{base.anything}` in `url_pattern` is a raw `AttributeError` traceback**, from the flag and from the file alike; `check_url_pattern` catches `KeyError`, `IndexError` and `ValueError` but not this fourth shape. Enumerating the fields with `string.Formatter().parse()` and rejecting anything outside the four names, or carrying an attribute or index suffix, closes both this and the `{base.__class__.__mro__}` rendering.
- **"Every field is always present" does not hold for exported properties.** A case with no `<property name="Requirement">` gets no `Requirement` key, while the build writes `Requirement: None`; the module docstring promises a schema "can simply require a field".
- **A renameable field can be pointed at a fixed field's name.** `file_option = "case"` makes the converter overwrite the case name with the report path, and the build crashes in `add_need` with "multiple values for keyword argument 'case'". The build half is pre-existing; with `fields.FIELDS` there is now one table to check against.
- **The CLI names the config file it used only under `-v`**, while the build logs it at INFO on every run. The output is cached bytes shaped by an implicitly discovered file up to three directories away; print it at least when the file was found outside the working directory.
- **Two doc sentences:** `cli.rst:206` still says `convert` for the table, and `cli.rst:176` describes the boundary as `.git` or `pyproject.toml`, which since #145 is only true outside a repository (measured on the uv-workspace layout: run from `packages/x/sub`, the root file is adopted).
- **Two documented behaviours have no test that would notice their loss.** The url_pattern check "before any report is read" (both tests pass a valid XML; pass a missing path with a bad pattern and assert the pattern error), and "reported once" for unexported properties (the fixture leaves out exactly one name, so once-per-run and once-per-name are indistinguishable). Moving the check late and firing once per name each survived the full suite.

### Nits

`normalise_remote_url("C:\\repos\\repo")` becomes `https://C/\repos\repo` (the scp pattern takes the drive letter as a host) where the docstring promises pass-through. `cli.rst:91` points at `needs_extra_links`, which sphinx-needs 8.5 deprecates for `needs_links`, so a `-W` project following the docs warns. The name-collision warning fires per need where every other diagnostic fires once per run. The body's "108 tests" is 119 by `--collect-only` (169 on master, 288 here).

Two behaviour deltas on the Sphinx side worth a line in the changelog rather than a fix: `result_text` and `remote_url` are now registered fields, so a project that already registered `remote_url` itself hits the swallowed `NeedsApiConfigWarning` branch. And the Sphinx 9 rendering of `InvalidConfigurationError` as a crash report is the pre-existing note from #145, now reachable by a `build.needs` typo too.

sphinxcontrib/test_reports/fields.py
```diff
@@ -0,0 +1,183 @@
...
+
+#: ``name -> (JSON type, description)`` for every field declared under a fixed
+#: name. The build registers all of them, on test-file, test-suite and
+#: test-case needs alike; the converter writes test-case needs and therefore
+#: declares :data:`CASE_FIELDS`.
+FIELDS: dict[str, tuple[str, str]] = {
+    "suite": ("string", "Test suite name"),
+    "case": ("string", "Test case name"),
+    "case_name": ("string", "Test case display name"),
+    "case_parameter": ("string", "Test case parameter"),
```
MaximilianSoerenPollak: What could a `case parameter` be? Do you have some examples?

sphinxcontrib/test_reports/identity.py
```diff
@@ -35,6 +35,24 @@ def short_hash(value: str, length: int = 5) -> str:
     return letters_only[:length].lower()
 
 
+#: ``name[param]`` as pytest spells a parameterised case.
+_PARAMETERISED = re.compile(r"^(?P<name>[^\[]+)($|\[(?P<param>.*)?\])")
```
MaximilianSoerenPollak: I'm sure this regex does more and less than what is wanted (as all regexes more complex than `[a-z]` or similar. 

But I guess since this is rather tightly defined by pytest it should be fine and do mostly what we wish for here.

sphinxcontrib/test_reports/identity.py
```diff
@@ -35,6 +35,24 @@ def short_hash(value: str, length: int = 5) -> str:
...
+
+
+def split_case_name(name: str) -> tuple[str, str]:
+    """``("test_x", "a-b")`` for ``"test_x[a-b]"``; the parameter is ``""``
+    when the name carries none.
```
MaximilianSoerenPollak: Test case name should always be filled no? 
I can't imagine a scenario where testcase name is empty. 
Testclassname yes, but casename I have not seen yet (at least in pytest)

sphinxcontrib/test_reports/needs_export.py
```diff
@@ -0,0 +1,358 @@
...
+NeedItem = dict[str, FieldValue]
+
+#: A parsed report: the path it was read from, and its top-level suites. The
+#: path is what the build's ``test-file``/``test-case`` directives record as
+#: the report-path field, so it travels with the suites.
+Report = tuple[str, Sequence[Mapping[str, object]]]
+
+#: Version key used when the caller does not supply one. needs.json requires a
+#: ``current_version``, but a converted report has no baseline history.
+DEFAULT_VERSION = "1.0"
```
MaximilianSoerenPollak, line 48: Dangerous to assume this, but I think for now and as a fall back this seems like a good idea to have.

sphinxcontrib/test_reports/needs_export.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def _literal_block(title: str, body: str) -> str:
+    """An RST literal block, indented so the need content stays valid.
+
+    The body is dedented as a whole, not line by line: XML pretty-printing adds
+    a common indentation that has to go, but a traceback or an assertion diff
+    is only readable if its *relative* indentation survives.
+    """
+    lines = textwrap.dedent(body).strip("\n").split("\n")
```
MaximilianSoerenPollak, line 75: Does this work, if you strip `\n' and then split by it, will you not just get the list with 1 item back?

sphinxcontrib/test_reports/needs_export.py
```diff
@@ -0,0 +1,358 @@
...
+    )
+    return f"\n**{title}**::\n\n{indented}\n"
+
+
+def build_content(case: Mapping[str, object]) -> str:
+    """Need content carrying the complete failure evidence.
+
+    googletest reports one ``<failure>`` per failed assertion, each with its own
+    message and body, plus captured output -- keeping only the first of them
+    loses exactly the detail a reader needs.
```
MaximilianSoerenPollak, line 87: I could see the usefulness in this, though I'm questioning if anybody ever would read the output of the CLI to asses the failures that they had in tests, like the actuall failure reason etc. 

Don't they just care about the result here? 
Or maybe I'm thinking wrong or too narrow about this

sphinxcontrib/test_reports/needs_export.py
```diff
@@ -0,0 +1,358 @@
+"""Turn parsed test reports into a needs.json payload.
```
MaximilianSoerenPollak, line 1: Unsure if dataclasses would make outputs / inputs more readable. 
I think it's fine as is as well but maybe in a future PR when one wants to cleanup it could be done.

sphinxcontrib/test_reports/remote.py
```diff
@@ -0,0 +1,106 @@
...
+#: The part after ``scheme://`` of an ``ssh://`` or ``git://`` remote:
+#: ``[user@]host[:port]/org/repo[.git]``. The port belongs to the transport
+#: endpoint, not to the web UI, so it must not survive into the browsable base
+#: -- and must not be mistaken for the first path segment.
+_SCHEME_REMOTE = re.compile(
+    r"^(?:[^@/]+@)?(?P<host>[^:/]+)(?::\d+)?/(?P<path>.+?)(?:\.git)?/?$"
+)
+
+#: ``git@host:org/repo.git`` and bare ``host/org/repo``.
+_SCP_STYLE = re.compile(r"^(?:[^@/]+@)?(?P<host>[^:/]+)[:/](?P<path>.+?)(?:\.git)?/?$")
```
MaximilianSoerenPollak, line 29: Can just repeat my comment from the regex above. 

They always have hidden issues, so I would avoid them wherever possible. 

But I guess we are trading this uncertainty here for brevity of the code, as if we would do the string mangling in python it would be 50+ lines easily. 
If that is a tradeoff you are comfy with in this repo that is fine by me.

sphinxcontrib/test_reports/remote.py
```diff
@@ -0,0 +1,106 @@
+"""Web URLs for test sources.
+
```
MaximilianSoerenPollak, line 2: What happens when the tests are not uploaded / have no remote , is the link a 404?

sphinxcontrib/test_reports/test_reports.py
```diff
@@ -334,24 +337,28 @@ def warn(message: str) -> None:
...
+    # back the same way (see `test_common.py`). One narrow ignore for the
+    # attachment; the rest of the function works on a typed mapping.
+    types: dict[str, list[str]] = getattr(app, "tr_types", None) or {}
+    app.tr_types = types  # type: ignore[attr-defined]
 
     # Collects the configured test-report node types
-    app.tr_types[app.config.tr_file[0]] = app.config.tr_file[1:]
-    app.tr_types[app.config.tr_suite[0]] = app.config.tr_suite[1:]
-    app.tr_types[app.config.tr_case[0]] = app.config.tr_case[1:]
+    types[app.config.tr_file[0]] = app.config.tr_file[1:]
```
MaximilianSoerenPollak, line 353: I remember from me trying to change this (add functionality to the need items (tr_file, suit, case etc.)  back in like early 2025 that it's quiet a hazzel. 

I don't think this is the scope of this PR but might be something to think about how one could 'easily' add their own thing or add things to the directives defined here. 

Or maybe it is already easily possible and I just did it really badly ^^ I also accept that :D

MaximilianSoerenPollak (commented):
Some comments here with questions. 

Also as I think stated in earlier review comments, need to ensure that no other option is overridden with what is newly defined here. 
For me it looks like those edgecases have been handled, but just to keep in mind.

sphinxcontrib/test_reports/needs_export.py
ubmarco: No, as `body` is a longer string with `\n` possibly at start, end or in the middle. `strip` only strips from start and end, not the middle ones. `split` then splits the middle `\n` into lines.

sphinxcontrib/test_reports/needs_export.py
ubmarco: I am pretty sure when thinking in a bigger scope, this is an information that the full traceability graph should have. AI can use that to identify what's wrong, just like CI failures are analysed by reading the logs.

ubmarco:
@chrisjsewell All points are addressed. The fixes are in six commits, `423e51b..d8940fc`. Every commit passes the full test suite and mypy on its own. At the head: 331 tests pass, pre-commit is clean, and the docs build without warnings.

Below is what changed, point by point. Each entry says what the problem was, what I changed, where the code is, and which test covers it.

### Blocking

**1. `--need-type` could get around the agreement rule.**
*Problem:* with a config file that has no `[test_reports.case]`, `--need-type check` was accepted. The build then dropped every imported need.
*Change:* whenever a config file is used, the CLI compares the final need type with the build's case type. If the file does not set `case`, the default `testcase` is used for the comparison, the same way the loader does it. Your example now stops with: `--need-type is 'check' but [test_reports] does not set case, so the build's test-case type is the default 'testcase'`. Without a config file (or with `--no-config`) the flag is free, as before.
*Where:* [`cli.py` line 478](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/cli.py#L478). *Tests:* `test_need_type_flag_without_a_case_entry_is_checked_against_the_default`, `test_need_type_flag_without_a_config_file_is_free`.

**2. `--extra-option` replaced the list from the file.**
*Change:* the flag now adds to the file's `extra_options` list instead of replacing it. To ignore the file, use `--no-config`. If a flag name is not in the file's list, it is still exported, but the CLI prints a warning, because the build only knows the fields from the file and `needimport` would drop the others. Example: `warning: --extra-option Requirement: not in the extra_options of [test_reports] in .../ubproject.toml ...`.
*Where:* [`cli.py` line 430](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/cli.py#L430). *Tests:* `TestExtraOptionFlag` (3 tests). `--link-property` still replaces the file's table, as documented; it is a mapping, not a list.

**3. The title was different from the directive's.**
*Change:* the converter now uses the plain case name as the title (`Addition`), exactly like the `test-case` directive. The IDs did not change.
*Where:* [`needs_export.py` line 218](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/needs_export.py#L218). *Test:* `test_type_and_title_match_the_directive_s`.

**4. `result` used a different word.**
*Change:* the `failure -> failed` mapping is removed. The converter writes `failure`, the same word the build writes. It is also the CSS class `tr_failure` and what the shipped report template filters on. One filter now works for imported and local test cases alike. The docs list the five possible values: `passed`, `failure`, `error`, `skipped`, `disabled`.
*Note for S-CORE:* their tooling will see `failure` instead of `failed`, and `error` as its own value instead of being folded into `failed`. Their metamodel accepts any string for `result`. What would need a small update there: the link colouring and a few `result == 'failed'` filters in docs pages.
*Where:* [`needs_export.py` line 233](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/needs_export.py#L233). *Test:* `TestResultVocabulary`.

**5. Renaming the fields only in `conf.py` silently puts data in the wrong field.**
*Change (docs):* `cli.rst` and `configuration.rst` now explain that the converter can only read `ubproject.toml`, what goes wrong when the renames are in `conf.py` only, and that the renames must be in `ubproject.toml`. The build-side warning you suggested is not in this PR. I can do it as a follow-up.

### Worth fixing

- **Credentials in the remote URL.** `normalise_remote_url` now removes `user:token@` from http(s) URLs. A test checks that the token is nowhere in the written file. [`remote.py` line 77](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/remote.py#L77).
- **A non-report XML gave an empty file without a word.** Every input file without test cases now prints a warning: `warning: pom.xml: no test cases found, so it adds no needs. A file that is not a test report parses as an empty one.` The exit code stays 0, because a real report can legitimately have zero cases. [`cli.py` line 278](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/cli.py#L278).
- **`{base.anything}` crashed with a traceback.** The URL pattern is now checked with `string.Formatter().parse()`. Only `{base}`, `{commit}`, `{file}` and `{line}` are allowed; anything else, including `{base.__class__}`, gives a clear error before any report is read. [`remote.py` line 86](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/remote.py#L86).
- **"Every field is always present" was not true for properties.** A property listed in `extra_options` is now written for every test case. If a case has no such property, the value is `null`, which is also what the build produces. [`needs_export.py` line 253](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/needs_export.py#L253).
- **A renamed field could take the name of a fixed field.** For example `file_option = "case"` crashed the build. There is now a list of reserved names, `fields.RESERVED_NAMES`. The loader checks `ubproject.toml` against it and the build checks `conf.py`. Both give a clear error. [`fields.py` line 113](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/fields.py#L113), [`projectconfig.py` line 431](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/projectconfig.py#L431), [`test_reports.py` line 378](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/test_reports.py#L378).
- **The config file was only named with `-v`.** When the CLI finds a config file by searching upwards, it now always prints `reading [test_reports] from /path/ubproject.toml`. `-v` still prints where an unsuccessful search stopped, and names a `--config` file. [`cli.py` line 336](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/cli.py#L336).
- **Two doc sentences.** Fixed: `build.needs` instead of `convert`; and the search stops at the repository root (`.git`), while `pyproject.toml` only counts outside a repository.
- **Two behaviours had no test.** Added: `test_the_pattern_is_checked_before_any_report_is_read` (a missing report plus a bad pattern: the pattern error comes first) and `test_left_out_properties_are_reported_once_for_all_names` (two names left out, one warning line).

### Nits

- `C:\repos\repo` and `/srv/git/repo` are passed through unchanged. [`remote.py` line 33](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/remote.py#L33).
- The docs point at `needs_links`, and mention `needs_extra_links` for sphinx-needs before 6.3.
- The name-collision warning is printed once per run, listing all names. [`needs_export.py` line 340](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/needs_export.py#L340).
- The PR description is updated: 162 new tests, 331 in total.
- Changelog: a note that `result_text` and `remote_url` are registered fields now, and a note that Sphinx 9 shows a config error as a crash report, also for a typo in `[test_reports.build.needs]`.

### One more fix

While answering a question about the `name[param]` regex, I found that it was not anchored at the end: `test_x[a]b` lost the `b`. Fixed in `d8940fc`, with 9 test cases. [`identity.py` line 42](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/identity.py#L42).

sphinxcontrib/test_reports/fields.py
ubmarco: pytest names a parametrised test like `test_add[1-2]`. We split that into two fields: `case_name = "test_add"` and `case_parameter = "1-2"`. A test without brackets has an empty `case_parameter`. I updated the field description so it says this, with the example.

ubmarco:
@MaximilianSoerenPollak The answers to your inline questions are in the threads.

To your summary point ("make sure that no other option is overridden with what is newly defined here"): this is now checked, not just assumed. `fields.RESERVED_NAMES` lists all fixed and core field names. If one of the three renameable fields (`file_option`, `source_file_option`, `source_line_option`) is set to such a name, the loader rejects the `ubproject.toml`, and the build rejects the same setting in `conf.py`. Both give a clear error instead of a crash. [`fields.py` line 113](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/fields.py#L113).

sphinxcontrib/test_reports/identity.py
ubmarco: Good instinct. It did a bit more than wanted: the pattern was not anchored at the end, so `test_x[a]b` became `("test_x", "a")` and the `b` was lost.

Fixed in `d8940fc`: the pattern now matches exactly pytest's `name[param]` shape and nothing else. A name that only happens to contain a bracket is kept as a whole. There are 9 test cases for it in `tests/test_identity.py`. Both the directive and the converter use this one function, so they can never split a name differently. [`identity.py` line 42](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/identity.py#L42)

sphinxcontrib/test_reports/identity.py
ubmarco: Yes, the name is always filled. The docstring was unclear: it meant that the *parameter* is empty when the name has no brackets. I reworded it: `("test_x", "a-b")` for `"test_x[a-b]"`, and `("test_x", "")` for `"test_x"`.

sphinxcontrib/test_reports/needs_export.py
ubmarco: Agreed, it is only a fallback. A needs.json must have a `current_version`, and a converted test report has no version of its own. You can set it with `version = "..."` in `[test_reports.build.needs]`, or with `--version`.

sphinxcontrib/test_reports/needs_export.py
ubmarco: Agreed, for a later PR. Both ends are plain dicts anyway (the parser output and the needs.json), so dataclasses would only help in the middle. #157 (renaming "extra options" to "fields") is a good moment to do that.

sphinxcontrib/test_reports/remote.py
ubmarco: Same trade-off, taken on purpose. The module docstring says that the full implementation (based on giturlparse) lives in sphinx-codelinks and should be shared once the extensions are merged.

Until then, tests pin 13 URL spellings: ssh with a port, `git://`, scp style, https with `.git`, plain `host/path`, `file://`, URLs with credentials, and Windows and Linux local paths. This review round also added the removal of credentials: GitLab's `CI_REPOSITORY_URL` contains the job token, and it was ending up in every need. [`remote.py` line 77](https://github.com/useblocks/sphinx-test-reports/blob/d8940fc45e8c94153297fe6e069dabf33e5217be/sphinxcontrib/test_reports/remote.py#L77)

sphinxcontrib/test_reports/remote.py
ubmarco: No link at all. Without both `--remote-url` and `--commit` (or the `remote_url` and `commit` keys in the file), the URL fields `external_url` and `remote_url` are empty strings, so nothing is rendered as a link. Giving only one of the two is rejected with an error. This is described under "Source links" in `cli.rst`.

sphinxcontrib/test_reports/test_reports.py
ubmarco: Not in scope for this PR, but noted. Since this branch, the fields of all three directives come from one table (`fields.FIELDS` and `fields.RENAMEABLE_FIELDS`). That is the place where an extension point would go. #157 tracks the field vocabulary.

chrisjsewell (approved):
Round two at `d8940fc4`, every round-one finding re-run by hand on the new head in the same three environments. All sixteen are addressed. Approving.

**Re-verified.** The eight-row need-type matrix: row 8 is now exit 2 with the message you quote, rows 1 to 7 unchanged. `--extra-option` is additive, the file's own `TestType` survives, and the warning names the flag, the key, the file and the remedy. Title and `result` match the directive, listed properties are present as `null` when absent, the token in a `--remote-url` appears nowhere in the artifact, `{base.__class__}` and friends are refused before any report is read, empty and non-report files warn per file, `RESERVED_NAMES` rejects `file_option = "case"` from both the loader and the build on both cells, and the two behaviours that had no test now have one each, red under my round-one mutations. The rebase is exact: the fifteen rewritten commits differ from the reviewed patch by one context line that #158 changed on master. 331 pass on both cells, mypy and the docs are clean, the Sphinx-free chain gained only `string` and `urllib.parse`, and my five mutations of the fix code were all caught.

**On the `split_case_name` fix**: no ID moves. All three ID routes hash the raw test-case name, proved by an artifact diff against the round-one head and by building and converting four odd name shapes. Only `case_name` and `case_parameter` change, for shapes pytest cannot produce, so nothing is breaking; it is a bugfix worth a changelog line all the same.

Four small things remain, none of them a reason to hold this. In or after this PR, your call:

- **The residue you named is real and the docs do not mention it.** With `tr_case` in `conf.py` alone and no `ubproject.toml`, a conversion with no flag imports as `testcase` and every need is dropped, silently on the converter's side; `--need-type` matching `conf.py` is the workaround. The new "renames belong in `ubproject.toml`" paragraph covers the three field names but not the case type. One sentence there closes the last silent route.
- **`RESERVED_NAMES` lists the need's fields, not `add_need`'s parameters.** `file_option = "need_type"` or `"docname"` still reaches the same `TypeError` in the build, and `"layout"` or `"template"` is accepted and silently turns the report path into a need option. `inspect.signature(add_need)` minus the current set is 24 names, identical on 8.5.0 and 6.0.1 (plus `allow_type_coercion` on 8.5); adding them as a literal list keeps `fields.py` Sphinx-free.
- The row-8 message ends with a stray comma before "in cfg.toml".
- The `case_name`/`case_parameter` change has no changelog line.

One observation for the record, not a finding: the artifact's `needs_schema` block is read by `needimport` only for defaults; `known_keys` comes from the project's own schema, so a field the build never registered is still dropped. The docs claim exactly that and no more.

PR #150, opened 2026-09-07
build: require sphinx-needs 6.0.1 and Sphinx 7.4

ubmarco:
Follows #147 (Python 3.11, merged). Split out so the version floors can be reviewed on their own;the declarative-config, plugin and converter PRs are rebased onto it and lose their remaining compatibility branches.

## Why

The package now requires Python 3.11. sphinx-needs raised its own floor to  ​3.10 with  ​6.0 (#1468 dropped 3.9) and requires Sphinx 7.4 from that release on. The versions this package still carried compatibility branches for — sphinx-needs 2.x to  ​5.x, Sphinx below  ​6.1 — are ones it no longer shares an interpreter range with in practice,and​ ​6.x is where field registration with a schema begins, which is what the extension relies on.

The exact floor is **6.0.1**:that is the first release whose `add_extra_option` takes `schema=` (sphinx-needs #1527). On 6.0.0 the call raised `TypeError`, which `_register_field` swallowed under a catch-all meant for "already registered" — so no field was registered and every directive failed with `InvalidNeedException`. The first CI run of this PR caught exactly that on the 6.0.0 cells.

## Changes

- `sphinx-needs>=6.0.1` and `sphinx>=7.4` (the floor sphinx-needs 6 implies anyway)
- nox and CI matrices: sphinx-needs 6.0.1 /​  ​6.3.0 /​  ​7.0.0 /​  ​8.0.0 × Sphinx  ​7.4.7 /​  ​8.1.3 (the Sphinx 5.0 cells could not have installed sphinx-needs 6)
- `sphinx_needs_update` registers fields with a schema unconditionally;the no-schema branch is gone
- `_register_field` catches only `NeedsApiConfigWarning` (a duplicate registration) instead of every exception,and no longer inspects `add_extra_option`'s signature
- the Sphinx version switches around `logging`, `status_iterator` and the test fixtures' path type are gone
- the `needs_schema_definitions` tests no longer skip
- the test fixtures' confs now switch to `needs_fields` at sphinx-needs 7.0.0 instead of 8..0.0:7.0.0 is where `needs_fields` was introduced and `needs_extra_options` deprecated (sphinx-needs #1611), so on  ​7.x the old option emits a `[needs.deprecated]` warning, which `test_test_file_needs_extra_options_no_warning` caught. Schema-based registration works since  ​6.0.1, so this only changes the 7.x path — no behaviour change for  ​6.x or​ ​ 8.x.

No behaviour change for any supported version. 110 tests pass on each of: sphinx-needs  ​6.​0.1 / Sphinx  ​7.4.7,​  ​7.0.0 / Sphinx  ​7.4.7,and​  ​7.0.0 / Sphinx  ​8.1.3;also verified on  ​8.5.0 / Sphinx  ​9.1.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: ["3.11", "3.12"]
         sphinx-version: ["7.4.7", "8.1.3"]
-        sphinx_needs-version: ["6.0.1", "6.3.0", "8.0.0"]
+        sphinx_needs-version: ["6.0.1", "6.3.0", "7.0.0", "8.0.0"]
     steps:
       - uses: actions/checkout@v2
       - name: Set Up Python ${{ matrix.python-version }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: ["3.11", "3.12"]
         sphinx-version: ["7.4.7", "8.1.3"]
-        sphinx_needs-version: ["6.0.0", "6.3.0", "8.0.0"]
+        sphinx_needs-version: ["6.0.1", "6.3.0", "8.0.0"]
     steps:
       - uses: actions/checkout@v2
       - name: Set Up Python ${{ matrix.python-version }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,8 +8,8 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.11", "3.12"]
-        sphinx-version: ["5.0", "8.1.3"]
-        sphinx_needs-version: ["2.1", "4.2", "5.1", "6.3.0", "8.0.0"]
+        sphinx-version: ["7.4.7", "8.1.3"]
+        sphinx_needs-version: ["6.0.0", "6.3.0", "8.0.0"]
     steps:
       - uses: actions/checkout@v2
       - name: Set Up Python ${{ matrix.python-version }}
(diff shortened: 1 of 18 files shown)


chrisjsewell (commented):
Checked the body's claims against sphinx-needs' history rather than taking them as read — all hold:

- `add_extra_option` at tag `6.0.0` has `description=` only; `6.0.1` adds `schema=` (`sphinx_needs/api/configuration.py`). So 6.0.1 is the true floor, and the unconditional `description=…, schema=…` call is valid across the whole supported range.
- sphinx-needs `6.0.0`'s `pyproject.toml` declares `sphinx>=7.4,<9`, so `sphinx>=7.4` here adds nothing the resolver would not already have forced.
- The narrowed handler is the change that could have bitten, and it is right at both ends: a duplicate registration raises `NeedsApiConfigWarning` from `NeedsSphinxConfig.add_extra_option` at `6.0.1` and from `NeedsSphinxConfig.add_field` at `8.5.0` (both the "already registered" and the "core field name" cases). Anything else now surfaces, which is what the first CI run of this PR proved the old catch-all was hiding.
- `needs_fields` is introduced and `needs_extra_options` deprecated in the `7.0.0` changelog (#1611), so moving the fixtures' switch from 8.0.0 to 7.0.0 is the correct threshold.
- `sphinx.util.display.status_iterator` (6.1+) and the pathlib-only `sphinx.testing` (7.2+) are both below the new floor.

Two non-blocking notes for a follow-up, not this PR:

- `packaging` is no longer imported anywhere under `sphinxcontrib/` after this — only `docs/conf.py` and the test fixtures use it. It is a dead runtime dependency now (harmless, sphinx pulls it in anyway).
- The matrix tops out at sphinx-needs 8.0.0, so the `add_field` branch (≥ 8.5) is only covered by the local run mentioned in the body. Worth an 8.5.0 cell, since that branch is the one every new user hits.

Happy for this to merge.

chrisjsewell (approved):

PR #154, opened 2026-09-08
build: drop dead `packaging` runtime dependency

### LINKED ISSUE #152: Remove dead `packaging` runtime dependency
Follow-up from [PR #150 review](https://github.com/useblocks/sphinx-test-reports/pull/150#pullrequestreview-5130337244) by @chrisjsewell.

Since PR #150 removed the last usages of `packaging` under `sphinxcontrib/`, it is no longer imported anywhere in the package — only `docs/conf.py` and the test fixtures use it. It is now a dead runtime dependency (harmless, since Sphinx pulls it in anyway, but misleading).

**Proposed:**
- Remove `packaging>=20.0` from the runtime `dependencies` in `pyproject.toml`.
- If still needed for docs/tests, move it to the appropriate dev/docs dependency group.

**Verification:**
- `grep -r packaging sphinxcontrib/` returns nothing (except pycache).
- Docs and tests still build/pass.

ubmarco:
Closes #152. Follow-up from @chrisjsewell's [review of #150](https://github.com/useblocks/sphinx-test-reports/pull/150#pullrequestreview-5130337244).

## What

`packaging` was declared as a runtime dependency, but PR #150 removed the last import of it under `sphinxcontrib/`. It is still genuinely used outside the package, so it moves to the extras that cover those consumers rather than being deleted outright:

| Consumer | Installed via | Extra it now lives in |
| --- | --- | --- |
| `docs/conf.py` | RTD (`.readthedocs.yaml`) and the `linkcheck` nox session | `docs` |
| 9 × `tests/doc_test/*/conf.py` | the `tests` nox session | `test` |

All of them use it for the same thing — the `Version(sphinx_needs.__version__) >= Version("7.0.0")` config gate.

The `>=20.0` floor is dropped rather than carried over: `packaging.version.Version` predates it by years, Sphinx already pins a much higher floor, and the surrounding extras entries only carry bounds where one actually matters.

## Verification

- `grep -r packaging sphinxcontrib/` → no matches (the issue's stated criterion).
- AST scan over all 18 modules under `sphinxcontrib/` for `import packaging` / `from packaging …` → **zero** hits, so this is not a grep artifact.
- Installed metadata after the change lists `packaging` only as `packaging ; extra == "docs"` and `packaging ; extra == "test"` — no longer a runtime requirement.
- `pytest -n auto tests/` → **169 passed**, unchanged from baseline (Python 3.12, Sphinx 8.1.3, sphinx-needs 8.0.0, from a fresh `.[test]` install).
- `sphinx -b html docs docs/_build/html` from a fresh `.[docs]` install → `build succeeded.`, no warnings.
- `pre-commit run --all-files` → all 7 hooks pass, `taplo-format` included.
- `uv sync --group dev && uv run mypy` → `Success: no issues found in 6 source files`.

No lockfile in the repo, so nothing to regenerate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,12 +23,13 @@ classifiers = [
 ]
 keywords = ["sphinx", "documentation", "test-reports"]
 requires-python = ">=3.11"
-dependencies = ["sphinx>=7.4", "lxml", "sphinx-needs>=6.0.1", "packaging>=20.0"]
+dependencies = ["sphinx>=7.4", "lxml", "sphinx-needs>=6.0.1"]
 
 [project.optional-dependencies]
 test = [
   "beautifulsoup4",
   "nox>=2025.2.9",
+  "packaging",
   "pytest-xdist",
   "pytest>=7.0",
   "pytest-cov",
@@ -36,6 +37,7 @@ test = [
   "sphinxcontrib-plantuml",
 ]
 docs = [
+  "packaging",
   "pillow",
   "sphinx-design",
   "sphinx-immaterial",


chrisjsewell (commented):
Checked by hand against the head, and it does what it says.

- `packaging` is imported nowhere under `sphinxcontrib/`; the only imports are the `Version(sphinx_needs.__version__) >= Version("7.0.0")` gate in `docs/conf.py` and in the test projects' `conf.py` files. Both consumers are installed through the extras the change moves it to: the `tests` nox session installs `.[test]`, the docs session and Read the Docs install `.[docs]`, and the coverage workflow uses `--all-extras`.
- That gate is still live, since the floor is sphinx-needs 6.0.1 and the branch flips at 7.0.0, so `packaging` cannot simply be dropped from those files yet.
- It is also safe by construction: Sphinx itself requires `packaging>=23.0`, so no install of this package has ever been without it. The change is about honest metadata rather than install behaviour, which is also why dropping the `>=20.0` floor costs nothing.

One count in the body: I find 8 test-project `conf.py` files importing it, not 9 (of 25 test projects). Worth fixing before the squash copies the body, but not a reason to hold the PR.

chrisjsewell (approved):

PR #159, opened 2026-09-10
build: make Sphinx and sphinx-needs the `sphinx` extra — bare package and `[pytest]` install without the toolchain

### LINKED ISSUE #155: Make Sphinx and sphinx-needs optional: the converter CLI and the pytest plugin run without them
## Problem

#148 (`test-reports build needs`) and #151 (the pytest plugin) are designed to run **without the documentation toolchain**, and they do: the CLI's import chain (`cli → junitparser / needs_export / identity / projectconfig / remote`) touches neither `sphinx`, `sphinx_needs` nor `docutils`, the plugin imports only `pytest`, and the package `__init__` already resolves `setup` lazily (PEP 562, #145) so importing a submodule never imports Sphinx.

The **install** does not match. `pyproject.toml` declares

```toml
dependencies = ["sphinx>=7.4", "lxml", "sphinx-needs>=6.0.1", "packaging>=20.0"]
```

so `pip install sphinx-test-reports` pulls **27 packages** (Sphinx, sphinx-needs, docutils, Jinja2, requests, …) into a test runner or a CI build action that only wants the converter or the decorator.

Verified in a venv holding only `lxml`, `packaging`, `pytest` and the package installed with `--no-deps`:

| | without Sphinx installed |
|---|---|
| `test-reports build needs` on the gtest and pytest sample XML | exit 0, valid `needs.json` envelope |
| the CLI's own tests (`test_cli_convert`, `test_cli_config`, `test_needs_export`) | 90 / 92 pass — the 2 failures are the tests that deliberately use the toolchain (`sphinx_needs.needsfile` schema check, `needimport` into a real build) |
| the plugin's own tests | 10 / 10 pass |
| a pytest run with `-p sphinxcontrib.test_reports.pytest_plugin` | `<testcase file="…" line="…">` and the link properties written, no `sphinx*`/`docutils` module loaded |

Two side findings:

- `packaging` is a dead runtime dependency: it is imported only by `docs/conf.py` and the test fixtures' `conf.py`, never by the package (added in #133, its user is gone).
- The existing "does not import Sphinx" tests scan `sys.modules` in a venv where Sphinx *is* installed. Nothing in CI exercises the case where it is absent, so a lazy `from sphinx_needs import …` inside a function would go unnoticed.

## Proposal

Make the documentation toolchain an extra; the core dependency is `lxml` alone.

```toml
dependencies = ["lxml"]

[project.optional-dependencies]
sphinx = ["sphinx>=7.4", "sphinx-needs>=6.0.1"]  # the Sphinx extension
pytest = ["pytest>=7.0"]                          # the plugin (#151)
```

- Docs: `pip install sphinx-test-reports[sphinx]` for the extension (`docs/install.rst`, README); the plain install for the CLI and the plugin.
- `[test]` gains `sphinx>=7.4` and `sphinx-needs>=6.0.1` (it currently inherits them from the core list); `[docs]` already lists them; the nox matrix pins both explicitly and is unaffected. Drop `packaging`.
- CI: a nox session that installs the package with `--no-deps` plus `lxml pytest` and runs the CLI and plugin tests (the two toolchain-dependent tests marked, e.g. `needs_sphinx`). This is the only thing that actually guarantees the property.

## The version-constraint gap

Extras are opt-in, so a Sphinx project that keeps doing `pip install sphinx-test-reports` (no `[sphinx]`) in an environment where Sphinx and sphinx-needs are *already installed* never shows the resolver the `>=7.4` / `>=6.0.1` floors. Today an old sphinx-needs gets upgraded at install time; after the change it stays, and the extension fails at build time instead.

There is no packaging mechanism that expresses "constrain X only if X happens to be present". What can be done:

1. **Enforce the floors at load time, with the install hint in the message.** `setup()` calls `app.require_sphinx((7, 4))` — Sphinx's own hook, raises `VersionRequirementError`. For sphinx-needs, the lazy `__getattr__` path (which already turns a *missing* sphinx-needs into a clean `ExtensionError`) also compares `importlib.metadata.version("sphinx-needs")` against the floor and raises `ExtensionError("sphinx-test-reports requires sphinx-needs>=6.0.1, found 5.1.0; install sphinx-test-reports[sphinx]")`. `packaging.version.Version` is available for the comparison because Sphinx itself requires `packaging>=23.0`. This turns the failure from a confusing traceback into an actionable message at the moment the extension is used — a mitigation, not a fix: nothing stops the mis-install itself.
2. **Keep the extension's hard dependencies and give the toolchain-free parts their own distribution** (`test-reports`: parser, CLI, plugin, depends on `lxml`; `sphinx-test-reports` depends on it plus Sphinx and sphinx-needs). The resolver then always sees the floors for extension users, and pytest-only users no longer install something classified `Framework :: Sphinx`. Costs: a second pyproject (flit builds one package per project), a second PyPI name, and the `sphinxcontrib.test_reports` import package cannot be shipped by two distributions — the core needs its own import package, with `sphinxcontrib.test_reports.cli` / `.pytest_plugin` kept as thin re-exports. Bigger, and can still be done later on top of the extras change.

Suggested path: the extras change plus load-time floor checks now, with a changelog entry that spells out the install-line change; revisit the split if the pytest-only audience grows.

## Open questions

- Extra name: `[sphinx]`, or something like `[extension]`?
- The install behaviour changes for existing users: ship in 1.5.0 with a prominent changelog note, or bump to 2.0?
- Go straight to the two-distribution layout instead?

Refs: #148, #151; #145 (lazy `setup`), #133 (where `packaging` came in).

ubmarco:
Closes #155.

## What changes

`pip install sphinx-test-reports` now installs `lxml` alone. Sphinx and sphinx-needs move to a `sphinx` extra:

```toml
dependencies = ["lxml"]

[project.optional-dependencies]
sphinx = ["sphinx>=7.4", "sphinx-needs>=6.0.1"]
pytest = ["pytest>=7.0"]   # from #151, unchanged
```

So the install line of a documentation project becomes `pip install "sphinx-test-reports[sphinx]"`, and a test runner or build action that only wants `test-reports build needs` no longer pulls the toolchain. The same goes for the pytest plugin #151 added as the `pytest` extra: `pip install "sphinx-test-reports[pytest]"` now yields the plugin, pytest and `lxml`, nothing of the toolchain. `[test]` gains the two floors it used to inherit from the core list; `[docs]` already had them; the nox matrix pins both explicitly and is unaffected.

The install page lists the three install lines side by side (`sphinx` extra for a documentation project, `pytest` extra for a test runner, bare package for a build action running the command), and the plugin's page, which said how to enable the plugin but not how to install it, names the extra.

## The version-constraint gap

An extra is opt-in, so a project that keeps installing the bare package next to an already-installed older toolchain never shows pip the floors. The lazy `setup` in `__init__.py` now enforces them at load time, before the extension is imported (an outdated sphinx-needs may well import and only fail later, inside a directive):

```
Extension error:
Could not load extension sphinxcontrib.test_reports: sphinx-needs 5.1.0 is installed, but sphinx-needs>=6.0.1 is required. Install the Sphinx extension's dependencies with: pip install "sphinx-test-reports[sphinx]"
```

A toolchain that is missing altogether keeps the existing "Could not import extension" path, which now names the extra too:

```
Extension error:
Could not import extension sphinxcontrib.test_reports; the Sphinx extension's dependencies are an extra, install them with: pip install "sphinx-test-reports[sphinx]" (exception: No module named 'sphinx_needs')
```

One deviation from the sketch in the issue: instead of `app.require_sphinx((7, 4))` plus a hard-coded sphinx-needs floor, the new `toolchain` module reads the `sphinx` extra's requirements from the package's own metadata (`importlib.metadata.requires`) and compares each installed version against its specifier. The floors then live in `pyproject.toml` alone -- #150 just moved them, and a copy in code would need a lockstep test -- and a test guards that the extra the check reads is the one `pyproject.toml` declares. `packaging` does the parsing; it is imported lazily and the check stands down without it, since it arrives with Sphinx (and pytest) wherever the extension can be loaded. A distribution that is *not installed* is deliberately left to the import: the import states it more precisely, and a toolchain importable from a source tree without metadata must not be refused.

On Sphinx 9 the error appears inside Sphinx's crash report, as the existing "Known" changelog entry already notes for the configuration errors; the note now covers this case too.

## Guaranteeing the property in CI

A `toolchain_free` nox session (Python 3.11 and 3.12, new CI job, wired into `all_good`) installs the package with the dependencies it declares -- `pip install ".[pytest]"`, no `--no-deps`, so the declared lists themselves are what is tested -- asserts that neither `sphinx`, `sphinx_needs` nor `docutils` is importable, and runs the converter's and the pytest plugin's test modules with `-m "not toolchain"`. The tests that need a build carry the new `toolchain` mark (the issue suggested `needs_sphinx`; `sphinx` itself is taken by Sphinx's own testing fixtures). `tests/conftest.py` loads `sphinx.testing.fixtures` only where Sphinx is importable, and one top-level import of the package's Sphinx-dependent `exceptions` module in `test_project_config.py` became local to the marked test that uses it.

#151's `plugin_floor` session (the plugin's tests on the oldest pytest per Python) installed `.[test]`, which after this change brings the toolchain back in. It now installs `.[pytest]` plus pytest-xdist and runs the same toolchain-absent guard, so the floor run doubles as the isolation proof on old pytest. It also forwards nox's posargs, which CI passes (`--full-trace`) and it ignored.

The mypy CI job installs with `uv sync --group dev --extra sphinx`; without the extra it would no longer see Sphinx and docutils.

## Open questions from the issue, as resolved here

- **Extra name:** `sphinx`, as proposed. It matches the `[docs]`/`[test]` style and the `[pytest]` extra from #151.
- **1.5.0 or 2.0:** not decided here. The changelog carries a "Breaking:" entry spelling out the install-line change; `version` is untouched, so the release can decide.
- **Two-distribution split:** not done, per the suggested path.
- **`[pytest]` extra:** on master since #151; this PR is what makes it toolchain-free, and the second commit proves and documents that.

## Verification

- `pytest -n auto tests/`: 407 passed (13 new in `tests/test_toolchain.py`)
- `nox -s toolchain_free-3.11 toolchain_free-3.12`: 308 passed, 3 skipped (the pytest-xdist tests; xdist is not installed there), 24 `toolchain`-marked deselected, in a venv holding the package, `lxml` and pytest only; the toolchain-absent guard passes
- `nox -s "plugin_floor(python='3.11', pytest_version='7.0.1')" "plugin_floor(python='3.12', pytest_version='7.3.2')"`: 63 passed each, toolchain absent
- `mypy`: clean (the new module is strict-checked, not excluded)
- `pre-commit run --all-files`: clean
- `sphinx-build -W` of the docs: clean

## Notes for the reviewer

- Rebased onto #151 (`pyproject.toml`, `noxfile.py`, `ci.yaml`, `tests/conftest.py` conflicted; all resolved as a union of both sides). The first commit is the original change; the second is the plugin-isolation follow-up and can be reviewed, or dropped, on its own.
- `docs/install.rst` also fixes two pre-existing typos on the page it rewrites ("must to be", "can be find").

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,6 +44,26 @@ jobs:
       - name: Run the plugin's tests on the oldest supported pytest
         run: nox --non-interactive --session "plugin_floor(python='${{ matrix.python-version }}', pytest_version='${{ matrix.pytest-version }}')" -- --full-trace
 
+  toolchain-free:
+    # The converter runs without Sphinx; this installs the bare package and
+    # runs its tests where Sphinx is absent, the only thing that guarantees it.
+    name: "Test without Sphinx (python: ${{ matrix.python-version }})"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.11", "3.12"]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set Up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Nox
+        run: |
+          python -m pip install nox
+      - name: Run Tests
+        run: nox --non-interactive --session "toolchain_free-${{ matrix.python-version }}" -- --full-trace
+
   pre-commit:
     name: pre-commit
     runs-on: ubuntu-latest
@@ -68,8 +88,10 @@ jobs:
         uses: astral-sh/setup-uv@v5
 
       - name: uv sync
+        # The toolchain is an extra: without it, mypy cannot follow the
+        # extension's Sphinx and docutils imports.
         run: |
-          uv sync --group dev
+          uv sync --group dev --extra sphinx
 
       - name: Run mypy
         run: |
@@ -99,6 +121,7 @@ jobs:
     needs:
       - tests
       - plugin-floor
+      - toolchain-free
       - pre-commit
       - mypy
       - linkcheck
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,8 +45,9 @@ jobs:
         run: nox --non-interactive --session "plugin_floor(python='${{ matrix.python-version }}', pytest_version='${{ matrix.pytest-version }}')" -- --full-trace
 
   toolchain-free:
-    # The converter runs without Sphinx; this installs the bare package and
-    # runs its tests where Sphinx is absent, the only thing that guarantees it.
+    # The converter and the pytest plugin run without Sphinx; this installs the
+    # package with the `pytest` extra alone and runs their tests where Sphinx is
+    # absent, the only thing that guarantees it.
     name: "Test without Sphinx (python: ${{ matrix.python-version }})"
     runs-on: ubuntu-latest
     strategy:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,14 +12,17 @@ Unreleased
   documentation project becomes ``pip install "sphinx-test-reports[sphinx]"``.
   The bare package brings only ``lxml``, the dependency of the ``test-reports``
   command, which runs in test runners and build actions that have no
-  documentation toolchain. An extra is opt-in, so the extension now checks the
+  documentation toolchain; the pytest plugin, the ``pytest`` extra, runs there
+  too. An extra is opt-in, so the extension now checks the
   installed toolchain against the versions the extra declares when Sphinx
   loads it: a missing or older Sphinx or Sphinx-Needs stops the build with a
   message naming the install line, instead of a traceback from inside a
   directive.
-* Testing: A CI job installs the bare package and runs the converter's tests
-  without Sphinx, so a toolchain import creeping into the command's import
-  chain -- or Sphinx creeping back into the dependency list -- fails the build.
+* Testing: CI installs the package with the ``pytest`` extra alone and runs the
+  converter's and the pytest plugin's tests without Sphinx -- on the newest
+  pytest and on the oldest the plugin supports -- so a toolchain import
+  creeping into either import chain, or Sphinx creeping back into a dependency
+  list, fails the build.
 * Feature: Support the googletest XML dialect: ``status="notrun"`` is reported
   as ``disabled`` instead of ``passed``, all ``<failure>``/``<skipped>`` parts
   of a test case are kept instead of only the first, ``RecordProperty`` values
@@ -61,7 +64,9 @@ Unreleased
   helper, for cases skipped at setup and under pytest-xdist too. Which
   properties exist, their XML names and which take lists is the
   ``test_reports_properties`` pytest ini option; S-CORE's model, which the
-  plugin was ported from, is the documented example. See :ref:`pytest_plugin`.
+  plugin was ported from, is the documented example. The plugin is the
+  ``pytest`` extra: ``pip install "sphinx-test-reports[pytest]"`` installs it
+  and pytest, without the documentation toolchain. See :ref:`pytest_plugin`.
 * Bugfix: ``tr_file_option``, ``tr_source_file_option`` and
   ``tr_source_line_option`` may no longer name a fixed field such as ``case``
   or ``result``, in ``conf.py`` or in the declarative file. The build
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,19 @@ Unreleased
 ----------
 :Released: under development
 
+* Breaking: ``pip install sphinx-test-reports`` no longer installs Sphinx and
+  Sphinx-Needs. They are the new ``sphinx`` extra, so the install line of a
+  documentation project becomes ``pip install "sphinx-test-reports[sphinx]"``.
+  The bare package brings only ``lxml``, the dependency of the ``test-reports``
+  command, which runs in test runners and build actions that have no
+  documentation toolchain. An extra is opt-in, so the extension now checks the
+  installed toolchain against the versions the extra declares when Sphinx
+  loads it: a missing or older Sphinx or Sphinx-Needs stops the build with a
+  message naming the install line, instead of a traceback from inside a
+  directive.
+* Testing: A CI job installs the bare package and runs the converter's tests
+  without Sphinx, so a toolchain import creeping into the command's import
+  chain -- or Sphinx creeping back into the dependency list -- fails the build.
 * Feature: Support the googletest XML dialect: ``status="notrun"`` is reported
   as ``disabled`` instead of ``passed``, all ``<failure>``/``<skipped>`` parts
   of a test case are kept instead of only the first, ``RecordProperty`` values
@@ -61,7 +74,8 @@ Unreleased
   file -- a wrong type, a rename onto a fixed field, a disagreeing need type
   -- as its crash report rather than as a one-line message; the message is in
   the report. A typo in ``[test_reports.build.needs]`` is reported the same
-  way.
+  way, and so is a missing or outdated toolchain refused when the extension
+  loads.
 * Support: Python 3.10 is no longer supported. It reached the end of upstream
   support, and dropping it lets the package read TOML with ``tomllib`` from the
   standard library instead of carrying a backport.
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -13,7 +13,10 @@ be cached by a build system, and the data is unavailable to anything else. The
 CLI splits the computation out; the documentation build only imports the result.
 
 The command imports no Sphinx code at all, so it can run as a build action in an
-environment that has no documentation toolchain installed.
+environment that has no documentation toolchain installed. Installed as
+``pip install sphinx-test-reports`` -- without the ``sphinx`` extra the
+:doc:`extension needs </install>` -- the package brings a single dependency,
+``lxml``.
 
 Converting a report
 -------------------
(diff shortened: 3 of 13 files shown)


chrisjsewell (approved):
Reviewed at `e3c5b277` by reproducing the claims: the three install lines from a built wheel in fresh environments, the load-time floor message with sphinx-needs downgraded under Sphinx 7.4.7 through 9.1.0, both new nox sessions run for real, the rebase diffed against the pre-rebase head, the suite on both cells, and six mutations of the new code. Approving. Two small things are worth landing before the tag, either here or in the release PR; they are at the end.

**What holds, measured.** The bare wheel installs the package and `lxml` and nothing else, and `test-reports build needs` converts a fixture with no `sphinx*` or `docutils*` module loaded. `[pytest]` adds pytest and its four dependencies, and the plugin writes properties from that environment; `[sphinx]` builds. The wheel's `Requires-Dist` is exactly the body's table. Both error messages are byte-exact against the body, the floor one with no `(exception:` suffix, and Sphinx surfaces it as the extension error it is, since it fetches `setup` with `getattr(mod, "setup", None)` and an `AttributeError` would have become "extension has no setup() function"; the mutation that raises `AttributeError` instead goes red on exactly that test. `packaging` is guaranteed at both ends of the range (`sphinx>=7.4` requires `>=23.0`; pytest `>=22`), and without it the check stands down as claimed. `toolchain_free` gives 308 passed, 3 skipped, 24 deselected on both interpreters, `plugin_floor` 63 on each, the CI logs agree, and the absent-toolchain guard fires when Sphinx is planted into that environment. The explicit nine-file list is necessary: a marker-only selection over `tests/` dies at import in `test_env.py` and `test_test_file.py` before any mark can deselect them. Every `toolchain` mark is on a test that needs a build, and dropping one from `TestImportIntoABuild` fails the session. The rebase is exactly what the body says: every non-conflicting file's patch is byte-identical to the pre-rebase commit and the changelog differs by one hunk offset. `uv sync --group dev --extra sphinx` plus mypy is clean, and without the extra it is 23 errors, so that CI change is necessary. 407 passed, mypy checks the new module, the docs build clean under `-W`.

### Worth landing before the tag

**1. The extra filter also keeps a marked core dependency.** `toolchain_requirements` keeps a requirement when its marker is true under `extra == "sphinx"`, and a marker that never mentions `extra` is true on its other clauses alone. Nothing in today's manifest has one, so this is latent, but `dependencies = ["lxml", 'tomli>=2 ; python_version < "3.11"']` is exactly the line this project would add if it ever lowered the floor again, and then a documentation build stops with "install `sphinx-test-reports[sphinx]`", which would not fix it. Reproduced through the test's own fake and through a real wheel with the line written into its installed `METADATA`. The predicate that is meant is "belongs to the extra": true with `extra="sphinx"` and false with `extra=""`, three tokens more. `test_other_extras_and_the_core_dependency_are_not_checked` could never have caught it, because its core dependency is unmarked and never reaches `evaluate`; one marked line in its `DECLARED` makes the test cover what its name promises.

**2. `[docs]` did not "already have them".** It has `sphinx` unpinned and `sphinx-needs>=6`, weaker than the extra's `7.4` and `6.0.1`, and `.readthedocs.yaml` installs `[docs]`, so Read the Docs resolves against the weaker pair and then meets the stricter load-time check. Today's resolver takes the newest of everything, so it does not bite; it is the same class of gap `[test]` was fixed for. One line: the two floors, or `sphinx-test-reports[sphinx]` in `docs`.

### At your discretion

- `prereleases=True` is load-bearing on packaging 23 to 25 and a no-op on 26, and there is no lockfile, so `test_a_prerelease_above_the_floor_is_accepted` passes with or without the flag on the packaging CI resolves; I confirmed by removing it. A case the flag changes on every packaging, or a direct assertion on the specifier, keeps the guard live.
- "On Sphinx 9 the error appears inside Sphinx's crash report" starts at Sphinx 8.2 (7.4.7, 8.0.2 and 8.1.3 print the one-line form; 8.2.3 and 9.1.0 the report), and the nox matrix tops out at 8.1.3, so CI never sees that shape.
- `hasattr(pkg, "setup")` and `getattr(pkg, "setup", None)` now raise on an outdated-but-importable toolchain where master returned; `dir`, `pydoc`, `inspect.getmembers`, `pkgutil` and every submodule import are unaffected because `dir()` does not list `setup`. Right for the one consumer that matters; a sentence in the `toolchain` docstring keeps the next maintainer from putting `AttributeError` back.
- `TestNoSphinxImport` in `test_cli_convert.py` checks `sphinx` and `sphinx_needs` but not `docutils`, unlike its sibling in the plugin's tests and the session guard.

**The release number.** `docs/install.rst` says `versionchanged:: 1.5.0` and `docs/pytest.rst` `versionadded:: 1.5.0`, while the body and #155 leave 1.5.0 versus 2.0 open and the changelog entry is labelled "Breaking:". This changelog has never carried that label in a released section; its one comparable change, requiring Sphinx-Needs 1.01 and later, shipped as 1.0.0. Whichever number is chosen, both directives have to say it, and nothing checks that they do.

chrisjsewell:
One addition after the approval, on the design rather than the code: I measured the version-constraint gap this PR's body describes and wrote up where I think the layout should end up once the package moves into the monorepo, in https://github.com/useblocks/sphinx-test-reports/issues/155#issuecomment-5625712982.

The short version: with this wheel, a uv project on the bare install line that pins `sphinx-needs==5.1.0` locks, syncs and passes `uv pip check` without complaint; only the `[sphinx]` extra makes the resolver refuse. So the load-time check here is the only guard, which is what the body says. In the workspace the two-distribution split this issue set aside becomes cheap, and it is the design in which the resolver enforces the floors again, so I would treat it as the destination and choose this release's version number with that in mind; the details and the timing trade-off are in the comment. Nothing in it changes the approval.

PR #160, opened 2026-09-10
🚀 Release v2.0.0

ubmarco:
## Release v2.0.0

Bumps the version, retitles the version directives that were written against
the planned 1.5.0, and turns the `Unreleased` changelog into the 2.0.0
section.

### Why 2.0.0 and not 1.5.0

Four changes in this cycle are breaking, so the major digit moves. The first
three break an existing install; the fourth breaks project content instead:

| Change | PR | What breaks |
| --- | --- | --- |
| Sphinx and Sphinx-Needs became the `sphinx` extra | #159 | A documentation project has to install `sphinx-test-reports[sphinx]`; the bare package no longer brings a toolchain |
| Python 3.10 dropped | #147 | 3.11 is the floor |
| Sphinx-Needs 6.0.1 / Sphinx 7.4 floors | #150 | The compatibility branches for older releases are gone |
| A failed test case's `result` is `failed`, not `failure` | #161 | A filter naming the value (`'failure' == result`), a custom `tr_report_template` — which carries two such filters from the shipped template it was copied from — and custom CSS on the generated `tr_failure` class, now `tr_failed` |

The docs were already written for 1.5.0, so every `versionadded` /
`versionchanged` directive naming it is retitled here.

### Changes

- **Version** → `2.0.0` in `pyproject.toml`, `docs/conf.py` (`version` =
  `2.0`, `release` = `2.0.0`) and `sphinxcontrib/test_reports/test_reports.py`
  (`VERSION`), matching the three places 1.4.0 touched.
- **Version directives** → `2.0.0` in `docs/cli.rst`, `docs/pytest.rst`,
  `docs/install.rst` and the five in `docs/configuration.rst`.
- **Changelog** → new `2.0.0` section, released 11.09.2026, with the
  `Unreleased` heading kept empty on top.

### Changelog notes

The entry prose is unchanged; it is regrouped by category the way the 1.4.0
section is, and each entry now carries its PR link. Two deliberate edits:

- The Python 3.10 drop and the Sphinx-Needs 6.0.1 / Sphinx 7.4 floors were
  labelled `Support:`. They are relabelled `Breaking:`, since they are what
  the major bump is for.
- #154 (`packaging` dropped as a runtime dependency, moved into the `docs` and
  `test` extras) had no changelog line. Added as a `Support:` entry — it
  changes the installed dependency set, which is this release's theme.

All 16 `Unreleased` entries survive; the new count is 17. Rebasing onto master
after #161 merged folded in its two entries — the `Breaking:` rename and the
`Improvement:` parser normalisation — bringing the section to 19. They are
placed by the same category grouping, and the section's opening prose gained
the rename as a fourth breaking change; its "Nothing else in an existing
project has to change" sentence had to be inverted, since the rename makes it
false.

### Verification

- `uv run pytest -n auto tests/` → **421 passed** (Python 3.14, from a fresh
  `uv sync --all-groups --all-extras`), re-run on the rebased branch; 407
  before #161, which added a test module.
- `uv run sphinx-build -b html docs docs/_build/html` → `build succeeded.`,
  no warnings, so the rewritten changelog is valid reStructuredText.
- `uv run pre-commit run --files <the 8 changed files>` → all hooks pass.
- `grep -rn "1\.5\.0" docs/ sphinxcontrib/ pyproject.toml` → no matches left.
- No test asserts the package version, so the bump needed no test changes.

### After merge

Push a `2.0.0` tag. `.github/workflows/release.yaml` triggers on tags matching
`[0-9].[0-9]+.[0-9]+`, which `2.0.0` does, and publishes to PyPI via trusted
publishing.

### One thing left for you to decide

`pyproject.toml` still classifies the project as
`Development Status :: 4 - Beta`. A 2.0.0 release is a natural moment to move
it to `5 - Production/Stable`, but that is a call about the project rather than
part of the version bump, so it is untouched here.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,48 @@ Unreleased
 ----------
 :Released: under development
 
+.. _`release:2.0.0`:
+
+2.0.0
+-----
+:Released: 11.09.2026
+
+A major release. ``pip install sphinx-test-reports`` no longer installs Sphinx
+and Sphinx-Needs, Python 3.10 is no longer supported, Sphinx-Needs 6.0.1 and
+Sphinx 7.4 are the oldest supported versions, and a failed test case's
+``result`` is spelled ``failed`` rather than ``failure``. What the bare package
+gains in return is a life outside a documentation build: a ``test-reports``
+command that turns test-result XML into a ``needs.json`` without running
+Sphinx, and a pytest plugin that writes the XML shape this extension reads. A
+project can also be described once, declaratively, in the ``[test_reports]``
+section of ``ubproject.toml`` instead of being restated in ``conf.py``.
+
+Upgrading a documentation project means adding the extra to its install line:
+``pip install "sphinx-test-reports[sphinx]"``. Beyond that, only a project that
+names the old ``failure`` result has to change anything: a filter on the value,
+a custom ``tr_report_template`` copied from the shipped one, and custom CSS on
+the ``tr_failure`` class.
+
+* Breaking: ``pip install sphinx-test-reports`` no longer installs Sphinx and
+  Sphinx-Needs. They are the new ``sphinx`` extra, so the install line of a
+  documentation project becomes ``pip install "sphinx-test-reports[sphinx]"``.
+  The bare package brings only ``lxml``, the dependency of the ``test-reports``
+  command, which runs in test runners and build actions that have no
+  documentation toolchain; the pytest plugin, the ``pytest`` extra, runs there
+  too. An extra is opt-in, so the extension now checks the installed toolchain
+  against the versions the extra declares when Sphinx loads it: a missing or
+  older Sphinx or Sphinx-Needs stops the build with a message naming the
+  install line, instead of a traceback from inside a directive.
+  `#159 <https://github.com/useblocks/sphinx-test-reports/pull/159>`_
+* Breaking: Python 3.10 is no longer supported. It reached the end of upstream
+  support, and dropping it lets the package read TOML with ``tomllib`` from the
+  standard library instead of carrying a backport.
+  `#147 <https://github.com/useblocks/sphinx-test-reports/pull/147>`_
+* Breaking: sphinx-needs 6.0.1 and Sphinx 7.4 are the oldest supported
+  versions. 6.0.1 is the first release whose ``add_extra_option`` takes a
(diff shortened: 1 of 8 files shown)


chrisjsewell (dismissed):
Checked at `5e634ee` on master `4e5acfe6` (#159). The bump is consistent in all three places (`pyproject.toml`, `VERSION` in `test_reports.py`, `docs/conf.py` as `2.0` / `2.0.0`), the eight `1.5.0` directives are retitled and no `1.5.0` is left anywhere (the three `sphinx-test-reports-1.4.0` strings in `tests/test_project_config.py` are fixture directory names, not versions), the 2.0.0 section follows the 1.4.0 section's shape, every one of the eleven PR links points at the PR whose title matches its entry, the sixteen entries survive with their prose and the #154 line is a fair addition, the relabelling of the 3.10 drop and the floors as "Breaking:" is what the major digit is for, the docs build clean under `-W`, and the suite is 407 passed. The tag pattern in `release.yaml` matches `2.0.0`, and that workflow is the one that published 1.4.0 on 2026-06-16, which PyPI confirms as the current release. 2.0.0 is the right number, for the reasons in the body and in #155. Approving.

Two edits before merging, no re-review needed:

- **The date.** The section says `:Released: 10.09.2026`; the tag is going out on the 11th at the earliest. The 1.4.0 section's date is the day its tag was pushed.
- **The body's last line.** The "Generated with Claude Code" footer is copied into the squash commit, and this history is about to be imported into `useblocks/sphinx-needs`, whose repository policy forbids those lines in commits, pull requests, issues and comments. Better to drop it here than to carry it across.

Three things for the record, none of them for this PR:

- The two items I named on #159 as worth landing before the tag, the extra filter that also keeps a marked core dependency and the weaker `[docs]` floors, are not on master and not in this PR. Both are latent today, so 2.0.0 is not wrong without them; they are a three-token and a one-line change, and a 2.0.1 or the monorepo move can carry them.
- The "Known" entry still says Sphinx 9 renders these errors as a crash report; measured on #159, the shape starts at 8.2.
- On the classifier: I would move to `5 - Production/Stable` with this release. A 2.0.0 that drops a Python version and raises floors is a project stating its support policy, which is what that classifier says.

chrisjsewell (dismissed):

chrisjsewell (approved):

Issue #114, opened 2025-04-01
Fix mypy warnings


ubmarco:
Started a PR already
https://github.com/useblocks/sphinx-test-reports/pull/113

chrisjsewell:
mypy is no longer used here. With the move into the Sphinx-Needs workspace the package is
type-checked by `ty`, configured once at the repository root, and — unlike the mypy configuration
this issue was about — it checks the whole package with no per-module exclude list. `uv run poe
typecheck` is green over `packages/sphinx-test-reports/src`. The `exclude` list that named this
issue no longer exists, so there is nothing to re-activate. Closing as obsolete.

sphinx-test-reports now lives in the Sphinx-Needs repository (`packages/sphinx-test-reports` of useblocks/sphinx-needs, since useblocks/sphinx-needs#1931); anything new goes there.

PR #161, opened 2026-09-11
feat(parsers)!: spell a failed result `failed`, not `failure` — one result vocabulary for both parsers

ubmarco:
## Why

S-CORE cannot adopt the `result` field as it is spelled today.

A test case's `result` was never a chosen name — it was whatever the input called it. The JUnit parser passed the name of the result-carrying `<testcase>` child through verbatim, so a failure came out as `failure`, after the `<failure>` element:

```python
RESULT_PART_KINDS = ("skipped", "failure", "error")   # XML element names
...
tc_dict["result"] = first_part["kind"]                # the element name, verbatim
```

The two states with no element of their own were invented in the `else` branch, and whoever wrote it reached for participles — `passed`, and `disabled` for googletest's `status="notrun"`. The JSON parser passed its report's own value through untouched.

Three consequences:

* the same product said `failure` for one case's result and `failed` for the count of them on its suite (`fields.FIELDS`), and `errors` next to `error`;
* the same logical outcome could arrive under two spellings depending on the input format, although `jsonparser`'s API is documented as being in sync with `junitparser`'s;
* `docs/directives/test_case.rst` documented the value as **`failed`** — which it never was. The page that documents the directive was simply wrong.

S-CORE's own colouring hook keys on pytest's vocabulary (`passed`/`failed`/`skipped`/`disabled`), so STR-converted needs would render uncoloured against it.

## What changed

Every state is now a participle: **`passed`, `failed`, `error`, `skipped`, `disabled`**.

`error` is deliberately *not* renamed. It agrees with the `errors` count beside it and with pytest, which spells that outcome `error` too. The pair that disagreed was `failure`/`failed`, and `failed` is the half that already appeared in a field name.

The substance is not the rename but the **normalisation step that did not exist**: the vocabulary is declared once, in a new `results` module, and both parsers map their input onto it. A JSON report written against the JUnit dialect no longer produces a different `result` than the XML it mirrors. A state this package does not know is still passed through untouched, so a `tr_json_mapping` pointing at a report with a vocabulary of its own keeps working.

The part-level `kind` stays the XML element name — it reports which element the evidence came from, and the converter capitalises it into the evidence heading ("Failure: …"). A test pins that divergence, and it is red-green verified: normalising `kind` too makes it fail.

The declared `result` description now names the states, so a consumer reading the declarations out of a produced `needs.json` — a schema check, a metamodel validator, S-CORE's tooling — learns the field's domain from the file:

```json
"description": "Test result status: passed, failed, error, skipped, disabled"
```

## Breaking — what a project has to update

1. **A filter naming the value**: `'failure' == result` becomes `'failed' == result`.
2. **A custom `tr_report_template`**: it carries *two* such filters from the shipped template it was copied from. This is the dangerous one — see below.
3. **Custom CSS** targeting the generated `tr_failure` class, now `tr_failed`. The stylesheet keeps rules for both, so the colours survive either way.

The value is also what `test-reports build needs` writes into `needs.json`, so a consumer of that file has to be updated with it. All of this is in the changelog entry.

## The untested coupling this surfaced

The shipped `test_report_template.txt` filters on the result value twice, and it had **no test coverage at all**: every fixture exercising `test-report` pointed `tr_report_template` at a template of its own. A rename could have emptied its "Failed test cases" table and count in every project using the default, silently.

A fixture now builds with the default template, and its count caught exactly that during this change — the parser started saying `failed` while the template still filtered `failure`, and the count went to zero.

## Verification

* `pytest -n auto tests/` — **421 passed**, 0 failures (baseline before the change: 407)
* `mypy` (strict) — **Success: no issues found in 14 source files**. The new module is type-checked rather than parked on the `#114` exclude list that both parsers sit on.
* `nox -s toolchain_free` — 3.11 and 3.12 both successful; the new tests run with Sphinx, sphinx-needs and docutils all absent, so the normalisation stays toolchain-free.
* `pre-commit run --all-files` — all hooks pass. The KOI8-encoded fixture template survived the edit (its test asserts on the decoded text).
* `sphinx-build docs` — clean. The documentation's own pytest example, filtering `'failed' == result` against a real report, renders **"Failed test cases: 4"** and emits `tr_failed` 12 times, so the filter and the CSS class are verified end to end rather than by inspection.

## Follow-ups, deliberately not in this PR

* `common.css` has no `tr_disabled` rule although the parser emits `disabled`, and no dark-theme variants at all.
* Annotating a *reference* to a test-case need with its result (the S-CORE feature this came from) — that belongs in sphinx-needs' `need_ref`/`meta_links`, with STR supplying the `tr_<result>` palette.

---

Targets `master`, adding to the `Unreleased` section; #160 gets rebased on top, which folds the entry into the 2.0.0 section.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,28 @@ Unreleased
 ----------
 :Released: under development
 
+* Breaking: a failed test case now carries the ``result`` value ``failed``
+  instead of ``failure``, so that every state is spelled the same way -- as a
+  participle, like the ``passed``, ``skipped`` and ``disabled`` beside it, and
+  like the ``failed`` count on a test-file and test-suite need. ``failure`` was
+  never a chosen name: it was the name of the JUnit ``<failure>`` element,
+  passed through by the parser. The full vocabulary is now ``passed``,
+  ``failed``, ``error``, ``skipped``, ``disabled``; ``error`` is unchanged,
+  because it agrees with the ``errors`` count beside it. **Three things in a
+  project have to be updated:** a filter naming the value
+  (``'failure' == result`` becomes ``'failed' == result``), a custom
+  ``tr_report_template``, which contains two such filters in the shipped
+  template it was copied from, and custom CSS targeting the generated
+  ``tr_failure`` class, which is now ``tr_failed`` (the stylesheet still
+  carries rules for both, so the colours survive either way). The value is also
+  what ``test-reports build needs`` writes into ``needs.json``, so a consumer
+  of that file -- a schema, a metamodel validator -- has to be updated with it.
+* Improvement: both parsers now map their input onto that one vocabulary
+  instead of each passing its own through, so a JSON report written against
+  the JUnit dialect no longer produces a different ``result`` than the XML it
+  mirrors. A state this package does not know is still passed through
+  untouched, so a ``tr_json_mapping`` pointing at a report with a vocabulary
+  of its own keeps working.
 * Breaking: ``pip install sphinx-test-reports`` no longer installs Sphinx and
   Sphinx-Needs. They are the new ``sphinx`` extra, so the install line of a
   documentation project becomes ``pip install "sphinx-test-reports[sphinx]"``.
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -41,7 +41,7 @@ lost evidence.
 Each test case becomes one need shaped exactly like the need the build's
 ``test-case`` directive creates for it: titled with the case name; ``suite``,
 ``case``, ``case_name``, ``case_parameter``, ``classname``, ``result`` (spelled
-as the build spells it -- ``passed``, ``failure``, ``error``, ``skipped``,
+as the build spells it -- ``passed``, ``failed``, ``error``, ``skipped``,
 ``disabled`` -- so one filter matches imported and local cases alike) and
 ``time``; the report path under ``file``; the test's source location under
 ``case_file`` and ``case_line``; a one-line ``result_text``; and the full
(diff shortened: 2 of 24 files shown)


chrisjsewell (approved):
Checked at `ce43520` on master `4e5acfe6`. Approving; one thing to drop from the body before merging, and the merge order matters for the release.

**Measured.** 421 passed, mypy clean with the new module in the strict set, the docs build clean under `-W`, and the new test module passes with no Sphinx installed. Three mutations, each reverted with a clean tree: the JUnit parser passing the element name through again fails seven tests across four modules; the shipped template filtering `'failure'` again while the parser says `failed` fails exactly one test, the new default-template fixture, which is the silent-count failure the body describes and which nothing else in the suite could see; and removing the JSON parser's normalisation fails the JSON vocabulary test. So the coupling this surfaced is now fenced, and the fence is not decorative.

**Completeness.** Every place the old value was named is covered: the shipped template's two filters, the docs' copy of it and the pytest example, the three fixture templates, the `test-case` directive page (which now lists the five values instead of the wrong example it carried), the CLI page, the plugin page, and the `result` declaration in the produced `needs.json`. The `test-results` directive's "Failures:" count comes from the suite attribute, not from case results, so it is unaffected; its rows now carry `tr_failed`, which the stylesheet serves alongside `tr_failure`. The three JSON fixtures still say `"result": "failure"`, correctly, since they are the input the normalisation is exercised on. The part-level `kind` stays the element name, pinned by a test, and the evidence heading still reads "Failure:". The one gap is your own follow-up, `tr_disabled` having no CSS rule.

**Before merging.** The body ends with a "Generated with Claude Code" footer; the squash commit copies it, and this history is about to be imported into `useblocks/sphinx-needs`, whose policy forbids those lines. Same note as on #160.

**Order.** This must land before #160, and #160 then needs the rebase you describe, so the "Breaking:" entry folds into the 2.0.0 section rather than starting a new Unreleased block after it; that rebase is also where the release date gets corrected.

